PR #104, opened 2015-02-12
http-client exposes module Network.HTTP.Client.Types

kmels:
Dear Michael,
This pull request permits access to the CJ constructor and the `expose` selector on `CJ` so that one can pick cookies in the cookieJar. In my case, in a redirect history, I am having one cookie with https off and another with http on, so I'd like to drop one for now..

### DIFF
--- a/http-client/http-client.cabal
+++ b/http-client/http-client.cabal
@@ -18,6 +18,7 @@ flag network-uri
 
 library
   exposed-modules:     Network.HTTP.Client
+                       Network.HTTP.Client.Types
                        Network.HTTP.Client.MultipartFormData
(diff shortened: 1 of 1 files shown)


snoyberg:
Doesn't the following function allow this?

http://www.stackage.org/haddock/nightly-2015-02-11/http-client-0.4.7.1/Network-HTTP-Client.html#v:destroyCookieJar

kmels:
That gets a list of cookies but not the CJ constructor?

snoyberg:
What about createCookieJar/

kmels:
Aha! That will do it! I'll close this then.

PR #105, opened 2015-02-12
Allow HTTPS proxies

jsl:
This commit fixes a bug where specifying a proxy prefixed by 'https' caused an HttpException to be raised. This seems like a scheme that should be valid, particularly since we're allowing an `https_proxy` to be specified in the environment.

With this change I also added test cases for previous behavior that wasn't tested, i.e., 'http:' should be a valid proxy uri scheme name, and 'ftp:' should not be a valid proxy uri scheme name.

### DIFF
--- a/http-client/http-client.cabal
+++ b/http-client/http-client.cabal
@@ -97,6 +97,7 @@ test-suite spec-nonet
   build-depends:       base
                      , http-client
                      , hspec
+                     , base-compat
                      , monad-control
                      , bytestring
                      , text
--- a/http-client/test-nonet/Network/HTTP/ClientSpec.hs
+++ b/http-client/test-nonet/Network/HTTP/ClientSpec.hs
@@ -1,17 +1,21 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Network.HTTP.ClientSpec where
 
-import           Control.Concurrent        (forkIO, threadDelay)
-import           Control.Concurrent.Async  (withAsync)
-import           Control.Exception         (bracket)
-import           Control.Monad             (forever, replicateM_)
+import           Control.Concurrent           (forkIO, threadDelay)
+import           Control.Concurrent.Async     (withAsync)
+import           Control.Exception            (bracket, try)
+import           Control.Monad                (forever, replicateM_)
+import           Data.Either                  (isRight)
 import           Network.HTTP.Client
-import           Network.HTTP.Types        (status413)
-import           Network.Socket            (sClose)
+import           Network.HTTP.Client.Internal (runProxyOverride)
+import           Network.HTTP.Types           (status413)
+import           Network.Socket               (sClose)
+import           System.Environment.Compat    (lookupEnv, setEnv, unsetEnv)
 import           Test.Hspec
-import qualified Data.Streaming.Network    as N
-import qualified Data.ByteString           as S
-import           Data.ByteString.Lazy.Char8 () -- orphan instance
+import qualified Data.Streaming.Network       as N
+import qualified Data.ByteString              as S
+import           Data.ByteString.Lazy.Char8   () -- orphan instance
+
 
 main :: IO ()
 main = hspec spec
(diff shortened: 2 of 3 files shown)


http-client/test-nonet/Network/HTTP/ClientSpec.hs
```diff
@@ -1,17 +1,21 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Network.HTTP.ClientSpec where
 
-import           Control.Concurrent        (forkIO, threadDelay)
```
jsl: Sorry for the diff in the import section - I'm mostly just keeping things lined up so there are some whitespace changes. The content changes in imports are:

* `import Data.Either (isRight)` added
* `import Network.HTTP.Client.Internal (runProxyOverride)` added -let me know if there is a way that you would prefer to test this functionality without depending on this internal function
* `import System.Environment.Compat (setEnv, lookupEnv, unsetEnv)` in order to replicate the failure that I experienced from inside the test suite. Again, let me know if there is a way that you'd prefer I test this functionality

http-client/http-client.cabal
```diff
@@ -97,6 +97,7 @@ test-suite spec-nonet
   build-depends:       base
                      , http-client
                      , hspec
+                     , base-compat
```
jsl: I'm adding this for the common functions across base versions to read and set environment variables (`lookupEnv`, `setEnv`, `unsetEnv`) so that I could reproduce the bug I found in the test suite. Let me know if there is a way that you think we can get around adding this import.

snoyberg:
Funny that you added the test for `ftp`... I just tested curl's behavior, and it seems to allow *any* scheme to be used (the one I happened to test was `ftp`). I think if we're going to make this change, we should allow any scheme. It's actually just as incorrect to specify https for a proxy as ftp IMO, since there's no SSL connection being made to the proxy server.

snoyberg:
To clarify: that's my only concern about this patch. Once we resolve that, I don't see a problem merging it in.

jsl:
@snoyberg I'm not sure what you mean when you say that specifying an https endpoint for a proxy is broken. It should work, with the caveat that it will break the certificate chain, since it substitutes the proxy server's certificate for the one at the ultimate destination. It's a relatively common thing to do, though, as there are several use cases for this sort of "MITM proxy."

Since the http_client libraries in this repo don't seem to support FTP, though, it makes sense to me that FTP URI schemes are rejected for these values since it would presumably cause an error when we actually reach out to the endpoint. In contrast, [Curl supports FTP](http://curl.haxx.se/) ([and it supports an FTP_PROXY variable as well](http://curl.haxx.se/libcurl/c/CURLOPT_PROXY.html)). I took a look at the Python and Ruby URI libraries, and they also support a `FTP_PROXY` environment variable in addition to HTTP_PROXY and HTTPS_PROXY ([ruby](http://ruby-doc.org/stdlib-1.8.7/libdoc/open-uri/rdoc/OpenURI.html), [python](https://docs.python.org/2/library/urllib.html)), but those libraries also support FTP so that makes sense to me.

It does make sense to me that in the current implementation of this library you can specify an http uri scheme in the `https_proxy` environment variable and an https uri scheme in the `http_proxy` endpoint and vice-versa to control whether or not to encrypt the communication from the original client to the proxy server.

I'd be happy to remove all filtering based on URI scheme, but just wanted to clarify the desired behavior here first. Please let me know how you would like to proceed.

Other than that, my only other nitpick related to the way we're handling these environment variables is that I think we should support upper-case versions, since I feel that's what most shell users expect, but that could be another issue/PR and I'd like to research what existing clients do before going ahead with implementation.

Thanks very much for writing this library, and for your review of this PR.

jsl:
Hi @snoyberg I looked into this more and I understand what you're saying about the proxying behavior for secure connections going over HTTP. I'm going to close this and re-open once I examine a bit more... perhaps it's not something that we should be fixing in this library at all.

Thank you!

PR #106, opened 2015-02-16
Provide a sane cross-platform way to check server certificates with openssl

redneb:
Currently `http-client-openssl` puts the burden of selecting the right trust roots to its users. If one wants to write an cross-platform client that verifies server certificates, there is no easy solution. On the other hand, `http-client-tls` does verify server certificates against the system's trust roots. This functionality is automatically offered by the `tls` package via the `x509-system` package. I wrote a new package [1],[2] that performs the same task for `HsOpenSSL`, using the same strategy as `x509-system`.

This pull request provides an easy cross-platform way to enable server certificate verification for `http-client-openssl` using `HsOpenSSL-x509-system`.

See also #2.

[1] https://hackage.haskell.org/package/HsOpenSSL-x509-system
[2] https://github.com/redneb/HsOpenSSL-x509-system

### DIFF
--- a/http-client-openssl/Network/HTTP/Client/OpenSSL.hs
+++ b/http-client-openssl/Network/HTTP/Client/OpenSSL.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 -- | Support for making connections via the OpenSSL library.
 module Network.HTTP.Client.OpenSSL
-    ( opensslManagerSettings
+    ( defaultOpensslManagerSettings
+    , opensslManagerSettings
     , defaultMakeContext
     , withOpenSSL
     ) where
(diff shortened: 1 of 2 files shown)


redneb:
Actually, I am afraid I have to withdraw this pull request. I just realized that it does not perform a complete verification of the server certificate.

Apparently, setting the `VerifyPeer` mode with `contextSetVerificationMode` in openssl only verifies that there is a trust chain starting from a trust root all the way to the peer's certificate. It does not check whether the validity period includes the current time or whether the hostname matches one of the names in the certificate (`commonName` or `subjectAltName`s). Apparently it is the application's responsibility to do those checks. So I will try to fix this and resubmit a pull request. One complication is that `HsOpenSSL` does not seem to provide a way to extract `subjectAltName` from a X.509 certificate.

PR #109, opened 2015-03-10
Add file-streamed request body utilities.

DanBurton:


http-client/Network/HTTP/Client/Types.hs
```diff
@@ -598,3 +599,11 @@ data ConnHost =
 -- specifying whether to use SSL.
 data ConnKey = ConnKey ConnHost Int S.ByteString Int Bool
     deriving (Eq, Show, Ord, T.Typeable)
+
+-- | Status of streaming a request body from a file.
```
snoyberg: Will also need a `Since` comment.

snoyberg:
Can you add a test case for this? I'm also thinking a little bit harder about whether the unix-compat dependency is acceptable here. I *think* it should be fine, but I seem to recall some issues with profiling versions segfaulting on Windows. Anyway, worst case scenario would be going back to your original approach which opened the file twice.

DanBurton:
I noticed the "open twice" pattern is used for MultiPartFormData:

```
size <- liftIO $ withBinaryFile f ReadMode hFileSize
```

https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/MultipartFormData.hs#L150

snoyberg:
Yeah, let's go for the open twice pattern and drop the unix-compat dep.

Only other thing is adding a test case for this function.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -57,6 +62,9 @@ import Data.Time.Clock
 import Control.Monad.Catch (MonadThrow, throwM)
 import Data.IORef
 
+import System.IO
```
snoyberg: Would be better to use an explicit import list or qualified import.

DanBurton:
Pending a few minor tweaks, then this PR should be ready to merge. I'll comment again when I've got those tweaks finished.

DanBurton:
Okay. The version bump is done, and the tmp directory thing now works on Windows. I ran the test suite on Windows and got the following failures:

http-client-test-failures

These errors are identical to when I run the snoyberg/master test suite, so I have presumably not introduced any regressions. The new test case I added is listed as passing.

stream-file-works

PTAL - should be ready to merge now.

snoyberg:
Thanks! Can you open up an issue about the Windows test suite failures? And heads up: I'll mostly likely ask you to look into those too (pinging @chrisdone).

DanBurton:
@snoyberg Done. I've opened up #111 about the Windows test suite failures.

Issue #107, opened 2015-02-25
Automatically call withSocketsDo everywhere


ndmitchell:
Currently, users must call `withSocketsDo` before using the http library. Likely that was because people assumed it to be expensive, or didn't know how it composed. We now know it isn't expensive, and the rules on how it composes are incredibly flexible: http://neilmitchell.blogspot.co.uk/2015/02/making-withsocketsdo-unnecessary.html

I suggest a call to `withSocketsDo` around each high level operation such as `withManager` (not sure if you're going to need it around `parseUrl`, but you should probably check). If you worry about the performance, I suggest you follow the CAF/unsafePerformIO/evaluate trick from the blog post above, but I doubt its going to be any noticeable performance on any version.

snoyberg:
Sounds good to me. I'll try to look at this some time next week.

DanBurton:
I realize that the motivation to continue providing `withSocketsDo` is for backwards compatibility, but at this point, shouldn't it be deprecated in favor of something like

```haskell
{-# INLINE initSockets #-}
initSockets :: IO ()
#if !defined(WITH_WINSOCK)
initSockets = return ()
#else
initSockets = evaluate withSocketsInit
```

Since we're no longer calling `shutdownWinSock`, there doesn't seem to be any point in retaining the "bracket" call pattern of `withSocketsDo`. If I'm not mistaken, all network activity in `http-client` requires a manager (either explicitly or under the hood), so it should be sufficient to `initSockets` sometime during `newManager`.

That said, I think it would be better to go with Simon Marlow's suggestion to have the initialization happen automatically and leave this stuff out of http-client entirely.

wrt people using old libraries but wanting to program in the new style of not worrying about `withSocketsDo` themselves, I suppose we could stick a `withSocketsDo` around `withManager`. However, I'm concerned: does the old `withSocketsDo` nest safely?

snoyberg:
> but at this point, shouldn't it be deprecated in favor of something like

But that's exactly the point of backwards compatibility: we *can't* deprecate it, use something else, and be compatible with preexisting versions in the wild. Many things might be nicer, but the fact is that we're stuck with a version of network out there that provides `withSocketsDo`, so we have to use it.

`withSocketsDo` can be nested safely; have a [look at the C functions](https://github.com/haskell/network/blob/v2.6.0.2/cbits/initWinSock.c).

ndmitchell:
[...]

PR #110, opened 2015-03-10
Gate manager creation with a withSocketsDo for Windows user conveninece.

DanBurton:
Here is an example I have tested on my Windows machine.

```haskell
-- example.hs
import Network.HTTP.Client

main = do
  withManager defaultManagerSettings $ \m -> do
    print =<< httpNoBody "http://google.com" m
```

When compiled against http-client-0.4.8 from hackage:

```
example.hs: FailedConnectionException2 "google.com" 80 False getAddrInfo: does n
ot exist (error 10093)
```

When compiled against this fork of http-client, it correctly retrieves and prints the Response.

Both times it was compiled against `network-2.4.2.3`, as included in the latest Haskell Platform.

(Tangent: both times, it spit out a bunch of warning messages about "X from blah32 is linked instead of __imp_X". These warnings do not cause it to crash, and do not prevent this sample use case from working correctly as far as I can tell.)

### DIFF
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -171,6 +171,7 @@ addToList now maxCount x l@(Cons _ currCount _ _)
 -- Since 0.1.0
 newManager :: ManagerSettings -> IO Manager
 newManager ms = do
+    NS.withSocketsDo $ return ()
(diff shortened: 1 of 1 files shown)


DanBurton:
As requested in #107.

snoyberg:
Looks good! Glad it turned out to be just the one location.

PR #114, opened 2015-04-06
Expect: 100-continue

romanb:
This PR proposes an implementation for the HTTP 1.1 `Expect: 100-continue` / `100 Continue` interaction whereby the client only sends the request body once it has received the `100 Continue`. The behaviour (including timeout behaviour) when `Expect: 100-continue` is **not** set should be identical to the way it is now. If the header is set, two timeouts apply, the first for waiting on the `100 Continue` and the second for waiting on the final response.

### Context / Motivation

For completeness, the specific situation where I ran into the need for this feature is a service that streams uploads from clients directly to S3. S3 closes idle connections quite aggressively after ~5 seconds. When these connections (in state `CLOSE_WAIT`) then end up being used for another upload before being closed by the `Manager`, part or all of the request body is pushed into the "dead" socket (buffer) and thus lost due to the source not being repeatable. That the connection is closed is usually not immediately discovered during sending, depending on how much data is being sent, but eventually fails quickly on receiving with `NoResponseDataReceived`. The subsequent retry on a new connection is then doomed to time out due to part of the data being gone already.
By expecting a `100 Continue`, the closed connection is guaranteed to be detected before the body is being sent. In the context of S3 it is also a good mechanism to handle [potential redirects](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTRedirect.html) without sending (parts of) the body twice. For these reasons, it seems to be a good idea to use `Expect: 100-continue` for streaming S3 uploads from non-repeatable sources.

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -89,9 +89,9 @@ httpRaw req0 m = do
     -- connections after accepting the request headers, so we need to check for
     -- exceptions in both.
     ex <- try $ do
-        requestBuilder req ci
+        cont <- requestBuilder req ci
 
-        getResponse connRelease timeout' req ci
+        getResponse connRelease timeout' req ci cont
 
     case (ex, isManaged) of
         -- Connection was reused, and might have been closed. Try again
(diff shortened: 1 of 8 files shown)


snoyberg:
The motivation and implementation both look good. I'm just reviewing a little bit further. Thank you for the contribution!

snoyberg:
I've reviewed and released this to Hackage as 0.4.10, thanks again!

PR #115, opened 2015-04-06
Allows the presence of 'content-length' header for chunked transfers.

lostbean:
After ed0b0b71f7bdd1a769a864bedd73c059cc1676e4 , the presence of both 'content-length' and 'transfer-encoding: chunked' rises an error. Although this is the ideal solution, there some real word, non-RFC 2616 servers out there that send chunkend bodies with 'content-length' headers.

This PR basically reverts ed0b0b71f7bdd1a769a864bedd73c059cc1676e4. The goal is to accept a chunked transfer even if the 'content-length' is present by ignoring its presence. In other words, the body acquisition should proceed as in any other chunked transfer.

### DIFF
--- a/http-client/Network/HTTP/Client/Types.hs
+++ b/http-client/Network/HTTP/Client/Types.hs
@@ -115,12 +115,11 @@ data HttpException = StatusCodeException Status ResponseHeaders CookieJar
                    -- ^ Environment name and value
                    --
                    -- Since 0.4.7
-
                    | ResponseLengthAndChunkingBothUsed
                    -- ^ Detect a case where both the @content-length@ header
-                   -- and @transfer-encoding: chunked@ are used.
+                   -- and @transfer-encoding: chunked@ are used. Since 0.4.8.
                    --
-                   -- Since 0.4.8
+                   -- Since 0.4.11 this exception isn't thrown anymore.
     deriving (Show, T.Typeable)
 instance Exception HttpException
 
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,3 +1,7 @@
+## 0.4.11
+
+* Ignore the 'Content-Length' header if the body contains chunked data [#115](https://github.com/snoyberg/http-client/pull/115)
+
 ## 0.4.10
 
 * Expect: 100-continue [#114](https://github.com/snoyberg/http-client/pull/114)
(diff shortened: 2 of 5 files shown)


lostbean:
I've noticed that the original test case uses an broken chunked string:

```haskell
        N.appWrite ad "HTTP/1.1 200 OK\r\ncontent-length: 0\r\ntransfer-encoding: chunked\r\n\r\n0"
```
The last (or unique) chunk of data has to be "0\r\n\r\n" (i.e. two CRLF). The original test makes the body reconstruction fails
```
*** Exception: recv: resource vanished (Connection reset by peer)
```
Although an exception is certainly expected in this situation, I would expect 'InvalidChunkHeaders' instead. Am I missing something here?

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -115,12 +115,6 @@ data HttpException = StatusCodeException Status ResponseHeaders CookieJar
                    -- ^ Environment name and value
                    --
                    -- Since 0.4.7
-
-                   | ResponseLengthAndChunkingBothUsed
```
snoyberg: We should leave this in for now, to avoid breaking code using it. We should also document that the exception will never be thrown.

snoyberg:
Looks good, I only had one comment. The reason no exception is thrown for the invalid chunking is similar to what you're doing here: I'm trying to be lenient in what we accept.

http-client/Network/HTTP/Client/Types.hs
lostbean: Sure, that is a valid point

lostbean:
I completely agree with being lenient, but my observation is about something else. Removing "\r\n\r\n" from the last chunk (as in the original test case) rises an error in the current implementation of `makeChunkedReader`, but IMHO it is the wrong type of error. We should do one of the following:

   1) recognize that this is the last chunk if the ckunk header is "0" or "0/rxxxxx" and then return the body without any error.
   2) throw `InvalidChunkHeaders` or `IncompleteHeaders` which would give a clear indication about the problem.

Apparently the function `recv` (`connectionRead` in `http-client`) is the one to blame here. It should return `empty` instead of rising an error when trying to read after the end of the stream. The documentation of `recv` says `For TCP sockets, a zero length return value means the peer has closed its half side of the connection. `

Does it make any sense ? :)

snoyberg:
I don't think the current behavior is incorrect, or even misleading. There's no real way to distinguish "the server purposefully sent an invalid chunk and then closed the connection" from "the connection was terminated before the end of the chunk header arrived." I don't see a problem with your approach either, but I'm not inclined to write code to make that change, and given that the change you're proposing would have some possibly far-reaching impacts on the rest of the codebase, I'd be hesitant to merge it in.

lostbean:
Indeed, both situations are indistinguishable and I understand that my propose might not worth to be implemented. I'm fine with this conclusion. I just wanted to bring this issue to light and have your opinion.

Thank you for merging this PR upstream!

snoyberg:
No problem, thank you for the PR!

PR #123, opened 2015-06-04
Fix getUri to insert "?" to uriQuery when necessary.

msakai:
`requestBuilder` function allows `queryString` of a `Request` not to begin with "?" character, and it inserts "?" when necessary.

`getUri` should do the same thing. Otherwise `applyCheckStatus` in `Network.Client.Core` module can produce `StatusCodeException` with wrong URI in its "X-Request-URL" field in such a case. I encountered the problem with influxdb package.


snoyberg:
Good catch, thank you!

PR #124, opened 2015-06-05
Support for auth via url

davidsiefert:
Example:
  http://jbond:007@httpbin.org/basic-auth/jbond/007

Does basic auth using jbond as username, 007 as password.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -28,7 +28,7 @@ module Network.HTTP.Client.Request
     ) where
 
 import Data.Int (Int64)
-import Data.Maybe (fromMaybe, isJust, fromJust)
+import Data.Maybe (fromMaybe, isJust)
 import Data.Monoid (mempty, mappend)
 import Data.String (IsString(..))
 import Data.Char (toLower)
@@ -114,15 +114,22 @@ getUri req = URI
     }
 
 applyAnyUriBasedAuth :: URI -> Request -> Request
-applyAnyUriBasedAuth uri =  applyBasicAuth (username authInfo) (password authInfo)
- where
-   authInfo = uriUserInfo $ fromJust $ uriAuthority uri
+applyAnyUriBasedAuth uri req =
+    if hasAuth
+      then applyBasicAuth (S8.pack theuser) (S8.pack thepass) req
+      else req
+  where
+    hasAuth = (notEmpty theuser) && (notEmpty thepass)
+    notEmpty = not . null
+    theuser = username authInfo
+    thepass = password authInfo
+    authInfo = uriUserInfo $ fromMaybe (URIAuth { uriUserInfo="", uriRegName="", uriPort="" }) $ uriAuthority uri
 
-username :: String -> S8.ByteString
-username = S8.pack . encode . takeWhile (/=':') . authPrefix
+username :: String -> String
+username = encode . takeWhile (/=':') . authPrefix
 
-password :: String -> S8.ByteString
-password = S8.pack . encode . takeWhile (/='@') . drop 1 . dropWhile (/=':')
+password :: String -> String
+password = encode . takeWhile (/='@') . drop 1 . dropWhile (/=':')
 
 encode :: String -> String
 encode = escapeURIString (not . isReserved)
@@ -258,7 +265,6 @@ browserDecompress = (/= "application/x-tar")
 --
 -- Since 0.1.0
 applyBasicAuth :: S.ByteString -> S.ByteString -> Request -> Request
-applyBasicAuth "" "" req = req
 applyBasicAuth user passwd req =
     req { requestHeaders = authHeader : requestHeaders req }
   where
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -111,16 +113,30 @@ getUri req = URI
     , uriFragment = ""
     }
 
+applyAnyUriBasedAuth :: URI -> Request -> Request
+applyAnyUriBasedAuth uri =  applyBasicAuth (username authInfo) (password authInfo)
+ where
+   authInfo = uriUserInfo $ fromJust $ uriAuthority uri
```
snoyberg: Can you explain why `fromJust` is being used here?

davidsiefert: Probably just out of convenience to get to a passing test as quick as possible, played in part by my Haskell skill level. :)  I know this will fail with an error on `Nothing`.  Perhaps I should use `... $ fromMaybe "" $ ...` or is there a better way?  Wondering about something like `do uinfo <- uriAuthority uri; authInfo <- uriUserInfo uinfo; applyBasicAuth ...` (along with whatever signature changes necessary to deal with Maybe monad operators).

snoyberg: I'm not familiar with this part of `network-uri`. Under what circumstances is `uriAuthority` a `Just` value, and when would it be `Nothing`?

davidsiefert: So @karun012

davidsiefert: So @karun012 and myself were looking through this `network-uri` code, and it seems that uriAuthority will never be Nothing....  here is some code we came across:
````haskell
uri :: URIParser URI                                                       
uri =                                                                      
    do  { us <- try uscheme                                                
        -- ; ua <- option Nothing ( do { try (string "//") ; uauthority } )
        -- ; up <- upath                                                   
        ; (ua,up) <- hierPart                                              
        ; uq <- option "" ( do { _ <- char '?' ; uquery    } )             
        ; uf <- option "" ( do { _ <- char '#' ; ufragment } )             
        ; return $ URI                                                     
            { uriScheme    = us                                            
            , uriAuthority = ua                                            
            , uriPath      = up                                            
            , uriQuery     = uq                                            
            , uriFragment  = uf                                            
            }                                                              
        }                                                                  
...
hierPart :: URIParser ((Maybe URIAuth),String)
hierPart =                                    
        do  { _ <- try (string "//")          
            ; ua <- uauthority                
            ; up <- pathAbEmpty               
            ; return (ua,up)                  
            }                                 
    <|> do  { up <- pathAbs                   
            ; return (Nothing,up)             
            }                                 
    <|> do  { up <- pathRootLess              
            ; return (Nothing,up)             
            }                                 
    <|> do  { return (Nothing,"")             
            }                                 
...
uauthority :: URIParser (Maybe URIAuth) 
uauthority =                            
    do  { uu <- option "" (try userinfo)
        ; uh <- host                    
        ; up <- option "" port          
        ; return $ Just $ URIAuth       
            { uriUserInfo = uu          
            , uriRegName  = uh          
            , uriPort     = up          
            }                           
        }                               
````
So I could be wrong, but it looks like it is always going to result in a `Just`....?

Comment (snoyberg) [inline http-client/Network/HTTP/Client/Request.hs]: That may be true today, but network-uri could change in the future without alerting us. I think your idea of `fromMaybe ""` could make sense here.

But I think I've got another problem with this: this code should only call applyBasicAuth if it actually finds a username and password, otherwise it's changing the behavior of existing programs incorrectly.

**Comment by davidsiefert** [comment]
Whoops! having a bad day--ignore that last commit. :)

**Comment by davidsiefert** [comment]
Closing for now, will resubmit.... :octopus:

**Comment by davidsiefert** [comment]
Okay, sorry I chickened out last time :chicken:
This should be working.  Is there a CI build that this runs through?

**Comment by snoyberg** [comment]
Huh, good catch, I didn't realize I hadn't turned on Travis here! I just did so now, but that won't help you for this PR. I can test on my system after reviewing.

File: http-client/Network/HTTP/Client/Request.hs [hunk 2]
```diff
@@ -111,16 +113,37 @@ getUri req = URI
...
+applyAnyUriBasedAuth uri req =
+    if hasAuth
+      then applyBasicAuth (S8.pack theuser) (S8.pack thepass) req
+      else req
+  where
+    hasAuth = (notEmpty theuser) && (notEmpty thepass)
+    notEmpty = not . null
+    theuser = username authInfo
+    thepass = password authInfo
+    authInfo = uriUserInfo $ fromMaybe (URIAuth { uriUserInfo="", uriRegName="", uriPort="" }) $ uriAuthority uri
```
snoyberg: How about `maybe "" uriUserInfo $ uriAuthority uri`?

davidsiefert: Ah, what a nice function.  Really cleaned it up. :smiley:

snoyberg:
Thanks!

davidsiefert:
Thanks guys for all the help and support for making this happen! :smile:

karun012:
:+1:

PR #128, opened 2015-06-30
treat empty string for http_proxy as equivalent to an unset environment variable

### LINKED ISSUE #127: system proxy prevents specifying a port with http-client?
Hi! To preface, my apologies if (from my networking inexperience) the following is actually a non-issue, or simply should be reported to a different project. I'm getting strange `FailedConnectionException2` exceptions when a system proxy is set. On OS X 10.10 Yosemite, I installed [`wreq`](http://hackage.haskell.org/package/wreq) using `lts-2.15` with stack. Network requests resolve *correctly* for requests on different ports, like 3000, when I **do not** have a system proxy configured (on my home machine):

```bash
$ stack ghci
Prelude> :m Network.Wreq
Prelude Network.Wreq> get "http://localhost:3000"
```
And if I don't have a server running, I get a sensible exception:

```
*** Exception: FailedConnectionException2 "localhost" 3000 False connect: does not exist (Connection refused
```

where port 3000 is given as parameter 2 of `FailedConnectionException2`. But under nearly the same conditions (the only difference being OSX 10.9 Mavericks on my work laptop), when I do have a system proxy set (`Apple → System Preferences → Network → $ACTIVE_INTERFACE → Advanced... → Proxies tab → Automatic Proxy Configuration → http://..../proxy.pac` ) on my network interface, requests using `http-client` do **not** resolve (strangely, as if they all try only port 80). So, even when I explicitly disable all my environment variables in my bash shell, like so:

```bash
$ http_proxy= https_proxy= HTTP_PROXY= HTTPS_PROXY= no_proxy= NO_PROXY= stack ghci
Prelude> :m Network.Wreq
Prelude Network.Wreq> get "http://localhost:3000"
```
I get the following incorrect exception: 

```
*** Exception: FailedConnectionException2 "localhost" 80 False connect: does not exist (Connection refused
```
where the exception is unexpectedly mentioning port **80** instead of 3000, even if

```
$ curl -v --noproxy localhost http://localhost:3000/
```

successfully resolves a locally hosted directory (with either `warp` or the traditional `python -m SimpleHTTPServer 3000`).

Any suggestions? If it helps to mention, my code which successfully ran last year (depending on `wreq == 0.2.0.0` in the cabal file) has stopped working on my system after recompiling, so I assume it's with the underlying dependency on `http-client` (`wreq-0.2.0.0` does not have any upper version-bound on `http-client` from what I see on hackage). Many thanks!

mmaz:
Resolves #127

### DIFF
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -486,11 +486,8 @@ envHelper :: Text -> EnvHelper -> IO (Request -> Request)
 envHelper name eh = do
     env <- getEnvironment
     case lookup (T.unpack name) env of
-        Nothing -> return $
-            case eh of
-                EHFromRequest -> id
-                EHNoProxy     -> \req -> req { proxy = Nothing }
-                EHUseProxy p  -> \req -> req { proxy = Just p  }
+        Nothing  -> return noEnvProxy
+        Just ""  -> return noEnvProxy
         Just str -> do
             let invalid = throwIO $ InvalidProxyEnvironmentVariable name (T.pack str)
             p <- maybe invalid return $ do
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks!

Issue #132, opened 2015-07-10
Proxy authentication does not work for HTTPS.


vasiliyl:
When connecting to HTTPS via proxy `Proxy-Authorization` header is not sent.
Proxy authentication for HTTPS should work the same way as for HTTP.

I'm behind the proxy with authentication right now, `https_proxy` environment variable is set to `http://user:pass@proxy:3128`, and here is what I get with stack: 

    $ stack build        
    Downloading nightly-2015-06-12 build plan ...ProxyConnectException "raw.githubusercontent.com" 443 (Left "StatusHeaders (Status {statusCode = 407, statusMessage = \"Proxy Authentication Required\"}) <...some more headers...>

Here is dump of stack's request:

    CONNECT raw.githubusercontent.com:443 HTTP/1.1

Here is dump of request done with curl which works fine:

    CONNECT raw.githubusercontent.com:443 HTTP/1.1\r\n
    Host: raw.githubusercontent.com:443\r\n
    Proxy-Authorization: Basic dXNlcjpwYXNz\r\n
    User-Agent: curl/7.42.1\r\n
    Proxy-Connection: Keep-Alive\r\n

snoyberg:
I'm not well set up to test this. Would you be able to take a stab at this? The change should be pretty trivial, just modify the `CONNECT` call in Network.HTTP.Client.Manager to respect the `proxy-authentication` header in `requestHeaders`.

vasiliyl:
I created merge request #133, it works for me. The code is trivial, but I'm new to haskell, please comment if there is a better solution.

snoyberg:
Merged to master, closing as resolved. If there are still issues, please report. Thanks for the PR!

PR #133, opened 2015-07-10
Support proxy authentication for CONNECT method #132

vasiliyl:


snoyberg:
One question: do we also need to remove the proxy-authentication header from the `requestHeaders` to prevent it from being sent to the destination server?

snoyberg:
Could you test out the vasiliyl-connect-proxy-auth branch and confirm that it still works? If so, I'll merge to master and release.

vasiliyl:
I am very sorry, I was out of town and could not respond. You are correct, that header should not be sent to destination server. Everything works as expected. Thank you for your work!

PR #134, opened 2015-07-10
history not being fully fetched

mwotton:


snoyberg:
Can you clarify what this PR is addressing?

mwotton:
apologies, I pushed this then realised the mistake I'd made - I thought I couldn't grab the full response from the history, but I'd called responseBody on it rather than brConsume, so i was only getting the first chunk. Meant to delete it when I realised but hadn't quite got round to it.

snoyberg:
Got it, thanks for clarifying :)

mwotton:
thanks for http-client, btw, it's made my life a lot easier:)

PR #139, opened 2015-07-25
Minor typos in DEPRECATED macros

tebello-thejane:
Although it is not entirely clear how one would go about making the suggested changes.


snoyberg:
Thanks

Issue #140, opened 2015-07-27
Support for NO_PROXY environment variable


hgiddens:
As @mmaz mentioned in #127, `NO_PROXY` is supported by various tools to avoid involving the proxy for a selection of hosts. For example, in the environment I'm using http-client in, external services are only accessible via the proxy, whereas internal services are only accessible directly. Currently I'm working around this explicitly, but direct support for the `NO_PROXY` environment variable would mean that everything Just Works™.

Other tools that support the variable include:
* [Emacs](http://www.gnu.org/software/emacs/manual/html_node/url/Proxies.html) (as @mmaz mentioned)
* [Wget](http://www.gnu.org/software/wget/manual/html_node/Proxies.html)
* [cURL](http://curl.haxx.se/docs/manual.html)
* [Python](https://docs.python.org/2/library/urllib.html)

Other systems typically support the same functionality, although not necessarily via environment variables or with the same syntax. For example, [Java sports a `http.nonProxyHosts` system property](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html).

If you're ok with the idea I'll have a go at an implementation.

snoyberg:
I am not familiar with the environment variable, but have no objection.
Pull request certainly welcome. Please try to include a link to relevant
specs in the comments.

snoyberg:
Merged in #145, closing. Thanks!

PR #145, opened 2015-08-07
Support no_proxy environment variable. #140

hgiddens:
Sorry if I've done any dumb stuff here. I tried to find a spec, and the nearest I could find was [some ancient info from W3C](http://www.w3.org/Daemon/User/Proxies/ProxyClients.html). I've thus implemented a subset of what other popular libraries support.

`no_proxy` contains a list of domain suffixes, separated by commas. The domain suffixes may optionally have leading periods; the commas may optionally have trailing spaces.

For example, if `no_proxy=internal.company.com, .other.company.com`, requests to `internal.company.com`, `foo.internal.company.com`, and `other.company.com` will avoid the proxy. Requests to `notactuallyinternal.company.com` will still go via the proxy.

###### Domain suffix vs. string suffix
wget and Python's urllib3 avoid the proxy if one of the `no_proxy` hosts is a string suffix of the request host, rather than a domain suffix (with special casing so `no_proxy=.foo.example` avoids the proxy for requests to `foo.example`).

###### Ignoring case/spaces
Some implementations ignore all spaces, rather than just `0x20`s following a comma. Some implementations also use case-sensitive domain handling.

###### Port-specific rules
Ruby's stdlib implementation can be made to apply only to specific ports.

###### Avoiding the proxy for CADR ranges
Python's urllib3 allows you to specify e.g. `no_proxy=internal.company.com,192.168.0.0/16`.

###### Avoiding the proxy for all requests
Ruby and cURL avoid the proxy for all requests if `no_proxy=*` is set.

### DIFF
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -37,6 +37,7 @@ import qualified Data.ByteString.Lazy as L
 
 import qualified Blaze.ByteString.Builder as Blaze
 
+import Data.Char (toLower)
 import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Text.Read (decimal)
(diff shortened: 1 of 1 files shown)


snoyberg:
Unless I'm misreading this, this code will require parsing the `no_proxy` environment variable from scratch on each request. It would be better to explicitly parse it immediately into some intermediate data structure, and then perform lookups on that.

hgiddens:
Yeah, you're quite right. Would `newManager` be an appropriate place for the parsing step?

snoyberg:
It's fine to do it in this function. Just do it outside of the inner function (in the same part of the code where `getEnvironment` is called).

hgiddens:
Is there anything else you'd like done on this one?

snoyberg:
I'll check now. For future reference: Github doesn't send notifications on new commits to a PR, so it's good to add a comment when making such a change.

snoyberg:
Thanks!

PR #147, opened 2015-08-09
move HasHttpManager from http-conduit to http-client

gregwebs:
relying on Travis to test this right now. Requires cabal file changes.


snoyberg:
Looks like a legitimate build failure. `HasHttpManager` should be exported from `Network.HTTP.Client` it seems.

snoyberg:
Also, if you rebase onto master first, you'll get the faster Travis builds.

PR #148, opened 2015-08-09
faster Travis build

gregwebs:
installing stack from source takes a long time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,24 +2,18 @@ language: haskell
 ghc:
   - '7.8'
 
-sudo: false
+sudo: true
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,24 +2,18 @@ language: haskell
 ghc:
   - '7.8'
 
-sudo: false
+sudo: true
```
snoyberg: This will disable the caching entirely. I just implemented a stack-based cached build here:

https://github.com/fpco/wai-middleware-crowd/blob/master/.travis.yml

Any objection to just using that approach?

gregwebs:
This is already a lot faster because it does not install stack from source. Hopefully the rest of the caching can get figured out to work properly.

.travis.yml
gregwebs: looks great

PR #151, opened 2015-08-15
raise minimum http-client version (needed for HasHttpManager)

intolerable:


snoyberg:
Good catch, thanks!

snoyberg:
Also edited .cabal file on Hackage.

PR #152, opened 2015-08-18
Expose brReadSome in Network.HTTP.Client

christian-marie:
I'm building a streaming interface to http-client (quiver-http) and am finding that brRead is often returning one to five byte chunks.

Quiver is nowhere near as optimised as conduit or pipes and I need to send larger chunks to get any performance.

Copy/pasting brReadSome and using it to build reasonable sized chunks made my performance problems go away entirely; would you consider exposing it?

### DIFF
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -154,6 +154,7 @@ module Network.HTTP.Client
       -- ** Response body
     , BodyReader
     , brRead
+    , brReadSome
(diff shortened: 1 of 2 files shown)


christian-marie:
Thanks. Hopefully the comment about the next release being 0.4.20 is prophetic enough.

snoyberg:
It's already on Hackage ;)

christian-marie:
<- impressed

PR #157, opened 2015-09-10
Add two examples to Haddocks

MaxGabriel:
Hey Michael, since http-client is pretty popular I wanted to add a couple examples to the Haddocks.


snoyberg:
Thanks!

PR #160, opened 2015-10-02
Prettify

ababkin:


snoyberg:
Please see the Travis build error. This looks like you may have changed semantics. I'll be honest: while the code is *prettier*, I don't think it's *clearer* than the explicit pattern matching.

snoyberg:
Closing in favor of #161

PR #161, opened 2015-10-04
Cleanup chain of case statements

jvans1:


snoyberg:
Out of curiosity, was this related at all to #160?

snoyberg:
The reason for the seemingly strange structure here is that, originally (back in tls-1.1.* at least), these exceptions were of different types. Making this change will break backwards compatibility with such old versions of tls. I'm OK with that, but we'll need a lower bound of 1.2 on the tls package in the .cabal file. Can you ping this PR when that change is made?

jvans1:
@snoyberg , I've changed the requirement on tls to >= 1.2. #160 is related, we found this code snippet and thought it would be fun to refactor independently and see what we each came up with. The previous structure makes sense given the different types, thanks for explaining. We were curious what lead to that code.

snoyberg:
Thanks!

PR #162, opened 2015-10-13
Feature request: Async streaming of request

christian-marie:
In trying to stream large amounts of data (hundreds of GB up, and down) over HTTP, I've hit an issue.

It appears that http-client only allows for the request to be streamed out, then the response streamed in, sequentially.

This is due to the continuation passing nature of the API, where you only get access to the response once the request has been fully sent.

My use case requires the output to be consumed as we go, as the server streams output before the request is fully consumed.

This PR is the approach I've personally taken to make this work, but I've ignored errors in the thread. I was wondering if you had input on alternative approaches, perhaps including modifying the API. If you are OK with this approach, then perhaps we can add exception handling and everything will be splendid.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -27,13 +27,14 @@ module Network.HTTP.Client.Request
     , password
     ) where
 
+import Control.Concurrent(forkIO)
 import Data.Int (Int64)
 import Data.Maybe (fromMaybe, isJust)
 import Data.Monoid (mempty, mappend)
 import Data.String (IsString(..))
 import Data.Char (toLower)
 import Control.Applicative ((<$>))
-import Control.Monad (when, unless)
+import Control.Monad (when, unless, void)
 import Numeric (showHex)
 
 import Data.Default.Class (Default (def))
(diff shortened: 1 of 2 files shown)


snoyberg:
I haven't tested this, but wouldn't the current API support this by having a separate thread write to a `Chan` or `MVar` and having the streaming constructor read from that value?

christian-marie:
I'm not sure how, based on the following:

1. writeStream (https://github.com/christian-marie/http-client/blob/christian/async-streaming/http-client/Network/HTTP/Client/Request.hs#L386) consumes the entire request stream in one loop.
1. requestBuilder (https://github.com/christian-marie/http-client/blob/christian/async-streaming/http-client/Network/HTTP/Client/Request.hs#L329) either eagerly evaluates writeStream (sendNow), or defers the writing until the headers are read (
https://github.com/christian-marie/http-client/blob/master/http-client/Network/HTTP/Client/Response.hs#L93), which is still before the response is returned to the user.
1. In order to consume the response, we must first evaluate httpRaw(https://github.com/christian-marie/http-client/blob/master/http-client/Network/HTTP/Client/Core.hs#L70), which by 1 and 2, fully consume the request stream.

snoyberg:
OK, now I get it. I'm very hesitant to include something like this. In Warp, we've had many corner cases around web standards regarding trying to read and write at the same time. HTTP is by its nature a half-duplex protocol, and this is attempting to bolt-on some full-duplex support.

christian-marie:
Correct, I would like to make it possible to implement full-duplex streaming clients using http-client.

We are using this to stream data about in a corporate environment, but @zadarnowski has just mentioned another interesting use case: supporting CONNECT.

We'll get back to you tomorrow with some convincing arguments.

christian-marie:
OK so this is my convincing argument: We've tested this usage of forkIO and it indeed appears to work.

If this isn't included in the library as a (experimental) feature, I'll have to duplicate http-client and include a http-client-async wherever I want to do full-duplex.

You're not convinced, I can tell, so If you definitely don't want to support full-duplex clients then would you consider adding an escape hatch for those of us that are less principled? Perhaps something like

```
  | RequestBodyHack (Request -> Connection -> (content length, send now, send later)

snoyberg:
It's true that I'm very much unconvinced, but worse than my lack of
convincing on the feature itself, this code looks unsafe to me. What's to
prevent a request from not being fully sent before a response is returned
(valid by the HTTP spec), and then another request sharing that connection
sending interleaved data over the same connection? There are a large number
of open questions along these lines.

christian-marie:
I'm not suggesting that this code is currently safe, but if we decided it's not an entirely insane idea then we might be able to address these questions.

christian-marie:
I will see if we can come up with a cleaner way to do this and get back to you tomorrow (again).

christian-marie:
Didn't get around to this today. Will close for now.

Issue #169, opened 2015-12-10
withResponseHistory: recv: invalid argument (Bad file descriptor)


dzhus:
I'm experiencing the following issue with http-client-0.4.24, on Mac OS X 10.10.5 and Ubuntu 14.04.

```haskell
import Prelude
import Network.HTTP.Client

followUrl' :: Manager
           -> String
           -> IO String
followUrl' mgr url = do
    req <- parseUrl url
    withResponseHistory req mgr $ \hr ->
        return $ show $ getUri $ hrFinalRequest hr
```

This fails with `Bad file descriptor` error for URLs which produce 404s:
```
λ> newManager defaultManagerSettings >>= (\m -> parseUrl "http://www.nytimes.com/2015/12/11/" >>= \r -> httpLbs r m)
*** Exception: StatusCodeException (Status {statusCode = 404, statusMessage = "Not Found"}) [("Server","Apache"),("X-App-Name","error"),...
λ> newManager defaultManagerSettings >>= (`followUrl'` "http://www.nytimes.com/2015/12/11/")
*** Exception: recv: invalid argument (Bad file descriptor)
```

Any ideas?

snoyberg:
Some initial information:

1. I can reproduce the problem
2. This doesn't seem to be a general problem with 404 responses, but rather specific to nytimes.com

snoyberg:
I've pushed a commit to master which should address the problem, can you give it a shot?

dzhus:
Thank you Michael, I think you've fixed it – the bug does not occur anymore.

Another URLs I've seen it happen with: 
- `http://dw.com/p/1HLgB?tw`
- `http://eonli.ne/1NProGo`
- `http://martha.ms/6011BTtHJ`
- `http://ss1.us/a/J09vNDCu`
- `http://uk.reuters.com/article/idUKKBN0TT18820151210`

snoyberg:
Cool, thanks for confirming. Uploaded to Hackage and closing.

Yuras:
Just saw it in the changelog on hackage. I think here we have bigger issue: we are trying to read from already closed socket, which is unsafe.

[...]

PR #170, opened 2015-12-12
Make sure we never read from or write to closed socket

Yuras:
Closed socket FD may be reused, so reading from or writing to it
is unsafe.

It is a fix I propose for #169

I removed the comment in `httpRedirect` because I don't think it is correct. `Bad file descriptor` occurs when the FD is closed locally, it has nothing to do with a remote side closing the connection. 

Also I added a separate `ConnectionClosed` exception instead of adding a constructor to `HttpException`. It is internal, so it should not be visible in API. I'll change it if you wish.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -85,15 +85,39 @@ makeConnection :: IO ByteString -- ^ read
                -> IO Connection
 makeConnection r w c = do
     istack <- newIORef []
+
+    -- it is necessary to make sure we never read from or write to
+    -- already closed connection.
+    closedVar <- newIORef False
+
     _ <- mkWeakIORef istack c
     return $! Connection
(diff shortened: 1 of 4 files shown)


http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -85,15 +86,35 @@ makeConnection :: IO ByteString -- ^ read
                -> IO Connection
 makeConnection r w c = do
     istack <- newIORef []
+
+    -- it is necessary to make sure we never read from or write to
+    -- already closed connection.
+    closedVar <- newMVar False
```
snoyberg: I'm concerned about adding `MVar` overhead for every operation. The way that the Manager system works, a `Connection` is only leased to a single thread at any given time, so therefore using an `IORef` should be cheaper and still thread-safe.

Yuras:
OK, I replaced `MVar` with `IORef`.

snoyberg:
Thanks!

PR #172, opened 2015-12-14
Derive ConnectionClosed from Typeable

sapek:
Fixes [build error on ghc 7.8](https://ci.appveyor.com/project/sapek/bond/build/1.0.506/job/d2lwmj1gut85f0k9#L381).

I don't know why this is not an error on 7.10.

PR #173, opened 2015-12-15
Add ghc 7.8 to Travis CI matrix

sapek:
Note that the ghc 7.8 build variant will fail because of the bug fixed with #172.


tmcgilchrist:
Just hit this issue. Is it worth adding links in the README to the passing/failing builds for master?

snoyberg:
Thanks!

PR #177, opened 2016-01-19
Enable managerModifyRequest to modify checkStatus

nikomi:
Here is the pull request to enable ```managerModifyRequest``` to modify ```checkStatus```.

In addition to the change to ```httpRaw``` a similar change to ```httpRedirect``` was required, plus a change to the function argument passed to ```httpRedirect``` which changed from ```Request -> IO (Response BodyReader, Maybe Request)``` to ```Request -> IO (Response BodyReader, Request, Bool)``` because it must always return the request potentially modified by ```httpRaw```.

JFI: we also noticed that ```managerModifyRequest``` will be called multiple times in the following circumstances:
* when a redirect is detected and followed (by calling ```httpRaw```)
* when a retryable error on a managed connection causes a retry (by calling ```responseOpen```)

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -22,6 +22,7 @@ import Network.HTTP.Client.Body
 import Network.HTTP.Client.Request
 import Network.HTTP.Client.Response
 import Network.HTTP.Client.Cookies
+import Data.Maybe (fromMaybe, isJust)
 import Data.Time
 import Control.Exception
 import qualified Data.ByteString as S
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client/Core.hs
```diff
@@ -69,11 +70,12 @@ httpLbs req man = withResponse req man $ \res -> do
 httpNoBody :: Request -> Manager -> IO (Response ())
 httpNoBody req man = withResponse req man $ return . void
 
+
 -- | Get a 'Response' without any redirect following.
 httpRaw
      :: Request
      -> Manager
-     -> IO (Response BodyReader)
+     -> IO (Request, Response BodyReader)
```
snoyberg: I'd misunderstood your previous comment, I thought you had intended to change the behavior of `httpRaw` but keep the type signature the same. Given this, making a new function with this type signature would be a better approach.

http-client/Network/HTTP/Client/Core.hs
```diff
@@ -205,31 +208,31 @@ applyCheckStatus req checkStatus' res =
 -- | Redirect loop
 httpRedirect
      :: Int -- ^ 'redirectCount'
-     -> (Request -> IO (Response BodyReader, Maybe Request)) -- ^ function which performs a request and returns a response, and possibly another request if there's a redirect.
+     -> (Request -> IO (Response BodyReader, Request, Bool)) -- ^ function which performs a request and returns a response, and possibly another request if there's a redirect.
```
snoyberg: Similar comment applies to this changed type signature

PR #179, opened 2016-01-20
Enable managerModifyRequest to modify checkStatus

nikomi:
Here is the new pull request to enable ```managerModifyRequest``` to modify ```checkStatus```.

Both ```httpRaw``` and ```httpRedirect``` have been retained as-is, even the callback API of ```httpRedirect``` is unchanged.

New functions ```httpRaw'``` and ```httpRedirect'``` (incl. an extended callback API) have been added.

JFI: we also noticed that managerModifyRequest will be called multiple times in the following circumstances:

when a redirect is detected and followed (by calling httpRaw)
when a retryable error on a managed connection causes a retry (by calling responseOpen)


snoyberg:
Thanks!

PR #180, opened 2016-02-21
Fix auth via url when password contain unsafe ASCII characters 

sifmelcara:
Before this fix, I cannot auth via url to access http://httpbin.org/basic-auth/user/p@ss
both `simpleHttp "http://user:p@ss@httpbin.org/basic-auth/user/p@ss"`
`simpleHttp http://user:p%40ss@httpbin.org/basic-auth/user/p@ss` will failed
(the former failed due to an exception (InvalidUrlException), the later failed authentication due to encoding p%40ss again).

After this fix, `simpleHttp "http://user:p%40ss@httpbin.org/basic-auth/user/p@ss"` works.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -23,8 +23,7 @@ module Network.HTTP.Client.Request
     , setQueryString
     , streamFile
     , observedStreamFile
-    , username
-    , password
+    , extractBasicAuthInfo
     ) where
 
 import Data.Int (Int64)
@@ -33,7 +32,7 @@ import Data.Monoid (mempty, mappend)
 import Data.String (IsString(..))
 import Data.Char (toLower)
 import Control.Applicative ((<$>))
-import Control.Monad (when, unless)
+import Control.Monad (when, unless, guard)
 import Numeric (showHex)
 
 import Data.Default.Class (Default (def))
@@ -47,7 +46,7 @@ import qualified Data.ByteString.Lazy as L
 import Data.ByteString.Lazy.Internal (defaultChunkSize)
 
 import qualified Network.HTTP.Types as W
-import Network.URI (URI (..), URIAuth (..), parseURI, relativeTo, escapeURIString, isAllowedInURI, isReserved)
+import Network.URI (URI (..), URIAuth (..), parseURI, relativeTo, escapeURIString, unEscapeString, isAllowedInURI, isReserved)
 
 import Control.Monad.IO.Class (liftIO)
 import Control.Exception (Exception, toException, throw, throwIO, IOException)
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -127,8 +127,8 @@ applyAnyUriBasedAuth uri req =
   where
     hasAuth = (notEmpty theuser) && (notEmpty thepass)
     notEmpty = not . null
-    theuser = username authInfo
-    thepass = password authInfo
+    theuser = unEscapeString . takeWhile (/=':') . authPrefix $ authInfo
+    thepass = unEscapeString . takeWhile (/='@') . drop 1 . dropWhile (/=':') $ authInfo
```
snoyberg: Why did you change these instead of modifying the implementation of `username` and `password`?

sifmelcara: Because I had not notice that `username` and `password` exported here are also used to apply basic auth.
I have update my commits.

sifmelcara:
I updated my commit and refactor the code a little bit.
(By adding a new function `extractBasicAuthInfo`)
`
 ghci> extractBasicAuthInfo $ fromJust $ parseURI "http://user:p%40ss@httpbin.org/user/p@ss"
Just ("user","p@ss")
`

By the way, this code should still work when user did not encode their username/password in URL and their username/password did not cause a parse error in parseUrl and did not contain '%'.

`
 ghci> extractBasicAuthInfo $ fromJust $ parseURI "http://u;ser:pa$$@httpbin.org/u;ser/pa$$"
Just ("u;ser","pa$$")
`

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -121,27 +120,21 @@ getUri req = URI
...
+    case extractBasicAuthInfo uri of
+        Just auth -> uncurry applyBasicAuth auth req
+        Nothing -> req
+
+-- | Extract basic access authentication info in URI.
+-- Return Nothing when there is no auth info in URI.
+extractBasicAuthInfo :: URI -> Maybe (S8.ByteString, S8.ByteString)
+extractBasicAuthInfo uri = do
+    userInfo <- uriUserInfo <$> uriAuthority uri
+    when (':' `notElem` userInfo) Nothing
```
snoyberg: More idiomatically:

```haskell
guard (':' `elem` userInfo)
```

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -121,27 +120,21 @@ getUri req = URI
...
+    let username = takeWhile (/=':') userInfo
```
snoyberg: It would be more efficient to use `let (username', password') = break (== ':') userInfo`

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -23,8 +23,7 @@ module Network.HTTP.Client.Request
     , setQueryString
     , streamFile
     , observedStreamFile
-    , username
```
snoyberg: This broke the test suite, please see the Travis build failure

sifmelcara:
Thank you. I updated my codes and test suite.

snoyberg:
Thanks!

Issue #185, opened 2016-04-08
Incorrect idle connection count in HTTP manager


nikomi:
Hi,

I believe the idle connection count is incremented incorrectly in module ```Network.HTTP.Client.Manager``` function ```putSocket``` line 156.

The code currently is
```
                let (l', mx) = addToList now (mMaxConns man) ci l
                    cnt' = idleCount + maybe 0 (const 1) mx
                    m' = ManagerOpen cnt' (Map.insert key l' m)
```
Since ```addToList``` returns ```Nothing``` if the connection _was_ added to the idle pool and ```Just ci``` if the connection was _not_ added (and has to be closed) the constants ```0``` and ```1``` should probably be reversed and the line should read:
```
                    cnt' = idleCount + maybe 1 (const 0) mx
```
What do you think?

snoyberg:
Good catch! Could you send a PR for this update?

snoyberg:
Resolved by #186, thanks!

PR #186, opened 2016-04-11
Fix incorrect idle connection count update in Manager

nikomi:
This is a fix for https://github.com/snoyberg/http-client/issues/185


snoyberg:
Thanks

PR #192, opened 2016-04-28
added missing Host: HTTP header for https CONNECT

dischoen:
Hi,

I usually live behind a company firewall. stack always had problems with our firewall, e.g:

```
$ stack templates
Failed to download templates. The HTTP error was: ProxyConnectException "api.github.com" 443 (Right (StatusCodeException (Status {statusCode = 400, statusMessage = "Bad Request"}) [] (CJ {expose = []})))

```

I found out that the "Host:" HTTP header was missing in the CONNECT request message of http-client and added that.
Now stack is working fine for me.

By the way, many thanks for stack itself. It gave back some motvation to continue with haskell...

Regards,
Dieter


dischoen:
I forgot: my proxy is an apache server with mod_proxy.
On the client i have set http_proxy and https_proxy env vars.

snoyberg:
LGTM, thanks!

PR #197, opened 2016-05-11
Fix http-client-tls link from pointing to http-client

MaxGabriel:
The change to https was just because that seemed to be the default for Stackage.org now


snoyberg:
Thanks!

Issue #193, opened 2016-04-29
Make adverse status codes not throw an exception


bitemyapp:
I get a lot of people complaining about http-client (usually via http-conduit and wreq) throwing exceptions for error-y status codes.

I think what Python's requests library is going to be less surprising to people:

```
>>> import requests
>>> r = requests.get('http://httpstat.us/401')
>>> r
<Response [401]>
>>> r.status_code
401
>>> r = requests.get('http://httpstat.us/500')
>>> r
<Response [500]>
```

I realize this is a major change, but as usual I think it's something where we're better off the sooner we get it fixed.

Yuras:
I'm sure you are aware of [checkStatus](http://hackage.haskell.org/package/http-client-0.4.28/docs/Network-HTTP-Client.html#v:checkStatus). Setting it to always return `Nothing` gives the behavior you what, doesn't it?

Not sure my opinion counts, but I think the default behavior (throwing an exception on non-2xx status) makes perfect sense, and I have a lot of code that relies on that.

bitemyapp:
@Yuras my point is about the default path. People coming to Haskell expecting "safety" find exceptions for a _status code_ being non-100/200/300 very surprising.

snoyberg:
As the docs mention, this is definitely a contentious issue, with people
having strong reasons on both sides. Breaking the API in a silent way in
such a situation is probably a bad idea.

bitemyapp:
@snoyberg no doubt. I do see the benefits of using exceptions for this, but we don't want surprise either way. Long as it's on your radar. I'll chew on it.

winterland1989:
Also mentioned in #23 , throwing on non-100/200/300 is okay depending on how you model HTTP protocol, the problem is that one may still want to access the respond body...etc inside the exception handler.

The way to do it with current API is to provide your own `checkStatus`, turn off throwing on non-100/200/300 ,  and check status code afterwards. so i suggest to provide a helper like:

```haskell
-- | turn off throwing on non-100/200/300 respond. 
silentBadStatus :: Request -> Request
silentBadStatus req = req{ checkStatus = \_ _ _ _ -> Nothing}
```

I also ask to move `checkStatus` from request level to manager level, i can't see a particular use case that per-requset granularity `checkStatus` is necessary, but i may miss something here, please discuss on this.

snoyberg:
I'm fine with making this a manager level setting as well, like we do with
timeouts. Want to make a PR?

[...]

PR #200, opened 2016-05-27
move status check to manager setting

winterland1989:
respond to #193 , disable status check is very common in many places, so it should be part of manager setting. The only problem is that this patch is a breaking change. @snoyberg please judge on this.

### DIFF
--- a/http-client/Network/HTTP/Client/Cookies.hs
+++ b/http-client/Network/HTTP/Client/Cookies.hs
@@ -47,7 +47,7 @@ isIpAddress =
             Just (i, x') | BS.null x' && i >= 0 && i < 256 -> go (rest - 1) y
             _ -> False
       where
-        (x, y') = BS.breakByte 46 bs -- period
+        (x, y') = BS.break (== 46) bs -- period
         y = BS.drop 1 y'
 
 -- | This corresponds to the subcomponent algorithm entitled \"Domain Matching\" detailed
--- a/http-client-openssl/test/Spec.hs
+++ b/http-client-openssl/test/Spec.hs
@@ -9,8 +9,9 @@ import qualified OpenSSL.Session       as SSL
 main :: IO ()
 main = withOpenSSL $ hspec $ do
     it "make a TLS connection" $ do
-        manager <- newManager $ opensslManagerSettings SSL.context
+        manager <- newManager $ (opensslManagerSettings SSL.context){
+                managerStatusCheck = \ _ _ _ -> Nothing
+            }
         withResponse "https://httpbin.org/status/418"
-            { checkStatus = \_ _ _ -> Nothing
-            } manager $ \res -> do
+            manager $ \res -> do
             responseStatus res `shouldBe` status418
--- a/http-client-tls/Network/HTTP/Client/TLS.hs
+++ b/http-client-tls/Network/HTTP/Client/TLS.hs
@@ -6,6 +6,7 @@
 module Network.HTTP.Client.TLS
     ( -- * Settings
       tlsManagerSettings
+    , tlsManagerSettings'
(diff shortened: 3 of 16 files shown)


snoyberg:
I'm -1 on this as a breaking change if it can be avoided (which I think it can be). It's still quite useful to be able to decide on a per request basis whether or not to throw an exception on non-200s, and obviously letting existing code continue to function is a nice thing.

winterland1989:
yes but i can't propose a good way with back compatibility to do this. in order to make old code continue to work, we have to add `checkStatus` field back with exactly same signature, and it's has to be able override manager setting(not other way around). but if we want to be able to express this overriding, the signature will have to be wrapped in a `Maybe`.

I'm happy to add back compatibility, but i need some input here.

snoyberg:
I'll need to think about it a bit, let me get back to you next week.

http-client/Network/HTTP/Client/Cookies.hs
```diff
@@ -47,7 +47,7 @@ isIpAddress =
             Just (i, x') | BS.null x' && i >= 0 && i < 256 -> go (rest - 1) y
             _ -> False
       where
-        (x, y') = BS.breakByte 46 bs -- period
+        (x, y') = BS.break (== 46) bs -- period
```
snoyberg: Can these kinds of orthogonal tweaks be included in a separate PR?

snoyberg:
Alright, let's start with the easier things to do to achieve future-compatibility when possible: instead of using accessors, we should be providing setter functions. So starting from the current version, we can add something like:

```haskell
setCheckStatus :: (Status -> ResponseHeaders -> CookieJar -> Maybe SomeException)
       -> Request -> Request
```

We can go further and deprecate the `checkStatus` accessor itself and recommend using `setCheckStatus`, which will make for a more graceful transition.

Now, as far as migrating towards having a manager-wide field with a request-level override. I can think of two approaches:

1. Change the type of `checkStatus` to be a `Maybe`, as you commented already. This is a breaking API change, but:
    * We retain all functionality of old versions
    * With the `setCheckStatus`, it will be trivial to write code that works with the old and new major version
2. A _really_ hacky trick: make the default value of `checkStatus` something like `\_ _ _ -> Just $ toException SpecialUnexportedTypeThatMeansUseTheDefaultHere`. That will technically allow us to express "please use the default from the manager" without any API change, though it will have a performance impact.

I'm not sure how important avoiding a breaking version number is here, and therefore whether we should do (2) or not. But one implementation strategy for all of this is:

1. Immediately add setters for all field accessors and deprecate the field accessors
2. Implement the hacky trick I mentioned
3. At some point in the future (after an announcement), we'll release a breaking API version of http-client that hides all of the field accessors and only exposes setters

winterland1989:
I think the transition is a little bit complex, i'm not sure about the cost/value ratio now. what about keeping current behavior, and move http-conduit's `setRequestIgnoreStatus` to http-client? maybe rename to something short like `ignoreBadStatus`.

snoyberg:
I'm fine with moving `setRequestIgnoreStatus` into http-client.

winterland1989:
Ok, i will come up another pr, i want to get these questions clear before doing it:

+ if this name `ignoreBadStatus` ok to you?

+ do you want me to include the `breakByte/break` fix? it's cause warnings on my ghc, i can seperate it with a clear commit message.

snoyberg:
I'd rather a name that is unambiguously a verb. The name suggested could be
taken as querying that ignore state.

I'm OK with the breakByte change, though I think the upstream warning was a
bad idea. I'd accept a PR for that change on its own, yes.

winterland1989:
close in favor of #201

PR #201, opened 2016-06-07
add setRequestIgnoreStatus

winterland1989:
let's add this since it's such a wanted function, i'll sent a second pr to fix the `breakByte` warning after you merge this.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -20,6 +20,7 @@ module Network.HTTP.Client.Request
     , needsGunzip
     , requestBuilder
     , useDefaultTimeout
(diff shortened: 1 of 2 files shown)


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -469,6 +470,13 @@ requestBuilder req Connection {..} = do
         <> fromByteString v
         <> fromByteString "\r\n"
 
+-- | Modify the request so that non-2XX status codes do not generate a runtime
+-- 'StatusCodeException'.
+--
+-- @since 0.4.28.1
```
snoyberg: Should be 0.4.29, I can fix that after merge

PR #203, opened 2016-06-09
Changed order of socket connect and socket tweak.

kosmoskatten:
Hi,

I've changed the order of connecting and tweaking sockets. My use case is to perform 'bind' on a client connection in order to bind it to a specific interface. Binding must be performed before connecting. For the setting of general socket options - which I guess is the main use case for the tweaking callback - this change should not matter, to the best of my understanding.

Cheers,
  --patrik

From commit comments:

When creating a Manager a callback can be registered using the
rawConnectionModifySocket helper function. The order when this
callback is called is changed. Now the callback is called before
the socket connection has opened. Some operations, e.g. a source
address bind, cannot be performed on connected sockets.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -165,8 +165,8 @@ openSocketConnectionSize tweakSocket chunksize hostAddress host port = do
             (NS.sClose)
             (\sock -> do
                 NS.setSocketOption sock NS.NoDelay 1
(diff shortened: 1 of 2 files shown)


snoyberg:
Looks fine to me. Would you mind also including a ChangeLog comment about this?

kosmoskatten:
Of course. I have added a little info to the ChangeLog.

snoyberg:
Thanks!

Issue #204, opened 2016-06-29
HTTP manager shared state corruption when sending wrong length for stream


sanketr:
Cross-post from [SO](http://stackoverflow.com/questions/38099279/http-manager-shared-state-corruption-when-sending-wrong-length-for-stream) where I asked about it.

Given a shared HTTP manager, it seems that if the requestBody is of type requestBodySource and if wrong length is supplied for the request body, then subsequent requests will crap out on same HTTP manager for about 20+ seconds (about 20 seconds in[ `aws`](https://www.stackage.org/lts-6.4/package/aws-0.13.2) package). There seems to be something about interaction of shared state and GivesPopper perhaps that is causing this issue. Here is a sample code that reproduces it - we use requestb.in for sending wrong length upload, and then try to read another valid URL on requestb.in.

```
{-# LANGUAGE OverloadedStrings #-}

    import           Data.Conduit.Binary (sourceFile)
    import           Network.HTTP.Conduit 
    import           Network.HTTP.Types
    import qualified Data.ByteString.Lazy as LBS
    import System.IO
    import Control.Monad.Trans.Resource (runResourceT)
    import Control.Concurrent.Async (async,waitCatch)
    import Control.Exception (displayException)

    main :: IO ()
    main = do
      {- Set up a ResourceT region with an available HTTP manager. -}
      httpmgr <- newManager tlsManagerSettings
      httpmgr2 <- newManager tlsManagerSettings
      let file ="out" -- some byte contents with length > 1
      lenb <- System.IO.withFile file ReadMode hFileSize
      let inbytes = sourceFile file 
      initReq <- parseUrl "http://requestb.in/saxbx3sa"
      putreq <- async $ runResourceT $ do
        let req = initReq { method = "POST",
          -- let us send wrong length in requestBodySource
          requestBody = (requestBodySource (fromIntegral $ lenb - 1) inbytes)}
        resp <- httpLbs req httpmgr 
        return (statusCode . responseStatus $ resp)
      putreqRes <- waitCatch putreq
      case putreqRes of
        Left e -> print $ displayException $ e
        Right r -> print $ r
      getreq <- async $ runResourceT $ do
        -- Let us do a GET on a different resource to see if it works
        initReq <- parseUrl "http://requestb.in/1l15sz21"
        let req = initReq { method = "GET"}
        resp <- httpLbs req httpmgr 
        return (statusCode . responseStatus $ resp)
      getreqRes <- waitCatch getreq
      case getreqRes of
[...]

PR #205, opened 2016-06-29
Added length validation for RequestBodyStream

sanketr:
This is proposed fix for #204. Tested with my code to verify that the error doesn't happen any more in case of wrong length, and that when the length is correct, data upload goes through.

I repurposed `writeStream` machinery and added `length` check for when chunking is false - not elegant solution :( 

Now, there is an additional branch - pseudo-code changes in bold below -
Before: if isChunked then return terminator else return ()
After: if isChunked then return terminator else **(if length error return 400 status code else return ())**

Not sure if Status code 400 is right one here. It seems to be appropriate for client-side error, to me.

If anything needs to be changed, please do let me know.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -396,9 +396,7 @@ requestBuilder req Connection {..} = do
                     if isChunked then connectionWrite "0\r\n\r\n"
                     -- If not chunked, then length argument is present
                     -- and should be validated
-                    else if ( fromJust mlen == n) 
-                             then return ()
-                             else throwIO $ WrongRequestBodyStreamSize (fromIntegral . fromJust $ mlen) (fromIntegral n)
+                    else unless (fromJust mlen == n) $ throwIO $ WrongRequestBodyStreamSize (fromIntegral . fromJust $ mlen) (fromIntegral n)
                 else do
                     connectionWrite $
                         if isChunked
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -396,9 +396,9 @@ requestBuilder req Connection {..} = do
                     if isChunked then connectionWrite "0\r\n\r\n"
                     -- If not chunked, then length argument is present
                     -- and should be validated
-                    else if ( fromJust mlen /= n) 
-                             then throwIO $ WrongRequestBodyStreamSize (fromIntegral . fromJust $ mlen) (fromIntegral n)
-                             else return ()
+                    else if ( fromJust mlen == n) 
+                             then return ()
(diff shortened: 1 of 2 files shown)


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -370,26 +370,32 @@ requestBuilder req Connection {..} = do
...
         return (Nothing, now, body)
     toTriple (RequestBodyIO mbody) = mbody >>= toTriple
 
-    writeStream isChunked withStream =
-        withStream loop
+    writeStream isChunked len withStream =
+        withStream (loop 0) 
       where
-        loop stream = do
+        loop n stream = do
```
snoyberg: This should probably be `!n` to avoid a space leak.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -370,26 +370,32 @@ requestBuilder req Connection {..} = do
...
     toTriple (RequestBodyStreamChunked stream) = do
-        let body = writeStream True stream
+        let body = writeStream True 0 stream
             now  = flushHeaders Nothing >> checkBadSend body
         return (Nothing, now, body)
     toTriple (RequestBodyIO mbody) = mbody >>= toTriple
 
-    writeStream isChunked withStream =
-        withStream loop
+    writeStream isChunked len withStream =
```
snoyberg: The `isChunked` and `len` is really providing redundant information. How about having a `mlen :: Maybe Int` parameter?

snoyberg:
I actually think it's a perfectly elegant solution. I put in some inline comments. My only other thought: this deserves its own exception constructor instead of using the `StatusCodeException`. How about `IncorrectRequestBodyStreamSize` or something like that?

sanketr:
Yep, agree it should be its own exception constructor. WrongRequestBodyStreamSize Word64 Word64 perhaps (same layout as ResponseBodyTooShort).

I have also applied all the changes that you suggested.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -370,26 +371,32 @@ requestBuilder req Connection {..} = do
...
+        isChunked = isNothing mlen
+        loop !n stream = do
             bs <- stream
             if S.null bs
-                then when isChunked $ connectionWrite "0\r\n\r\n"
+                then 
+                    if isChunked then connectionWrite "0\r\n\r\n"
+                    -- If not chunked, then length argument is present
+                    -- and should be validated
+                    else unless (fromJust mlen == n) $ throwIO $ WrongRequestBodyStreamSize (fromIntegral . fromJust $ mlen) (fromIntegral n)
```
snoyberg: I don't feel comfortable including a partial function like `fromJust` here. Why not just pattern match on `mlen` instead of using `isChunked`?

sanketr: Yep, agree. Removed partial function in latest commit.

PR #207, opened 2016-07-01
Negative SSL tests

ethercrow:
"expired" and "self-signed" ones pass while "wrong.host" doesn't.


snoyberg:
Are you planning on updating this PR for a fix for the failing tests?

ethercrow:
I was hoping for someone who knows how to fix this.

snoyberg:
@vincenthz Is checking if the SSL cert is for the wrong host something that should be checked by the connection package, or something that should be handled in http-client-tls?

vincenthz:
This is linked to not supporting public suffix list yet. For all practical purpose the badhost test doesn't make much sense IMHO. the wildcard is valid and used properly, but just happen not to be declared as a valid suffix in someone's list. https://github.com/vincenthz/hs-certificate/pull/52

vincenthz:
Also as soon, as this support is added, I'm sure people will start to open bugs for the opposite problem (aka "my own site with wildcard doesn't work even though the regex match")

vincenthz:
hm, it could be a wildcard bug afterall. This is seriously underspecified for a security spec :(

ocheron:
If interested I already worked on improving hostname matching in x509-validation: matching should be case-insensitive and wildcards should match only one level of subdomains.

vincenthz:
Yes, @ocheron this is definitely interesting.

ocheron:
OK so I sent vincenthz/hs-certificate#75.

oherrala:
I ran [TryTLS](https://github.com/ouspg/trytls) and this has been fixed in https://github.com/vincenthz/hs-certificate/pull/75:

```diff
- FAIL wrong hostname in certificate [reject wrong.host.badssl.com:443]
-      output: 200 OK
+ PASS wrong hostname in certificate [reject wrong.host.badssl.com:443]
+      output: HandshakeFailed (Error_Protocol ("certificate rejected: [NameMismatch \"wrong.host.badssl.com\"]",True,CertificateUnknown))

- FAIL invalid wildcard certificate Common Name [reject wildcard.mismatch.badtls.io:11007]
-      output: 200 OK
+ PASS invalid wildcard certificate Common Name [reject wildcard.mismatch.badtls.io:11007]
+      output: HandshakeFailed (Error_Protocol ("certificate rejected: [NameMismatch \"wildcard.mismatch.badtls.io\"]",True,CertificateUnknown))
```

ocheron:
@ethercrow Maybe you can add one more URL to test hostname case-insensitivity (something containing `BADSSL.COM` for instance)?

ethercrow:
Added case-insensitivity test and rebased.

ethercrow:
I don't understand why some Travis jobs have passed and the other have not.

ocheron:
I see several reasons:
- some tests for `http-client-openssl` fail because certificate validation is not enabled, so the TLS connection succeeds when it is expected to abort
- some jobs probably still use an older version of `x509-validation`

On the positive side, tests now succeed for `http-client-tls` with `x509-validation-1.6.5`.

http-client-openssl/test/Spec.hs
```diff
@@ -13,6 +13,26 @@ main = withOpenSSL $ hspec $ do
         withResponse (parseRequest_ "HEAD https://s3.amazonaws.com/hackage.fpcomplete.com/01-index.tar.gz") manager $ \res -> do
             responseStatus res `shouldBe` status200
             lookup "content-type" (responseHeaders res) `shouldBe` Just "application/x-gzip"
+
```
snoyberg: Any objection to removing these tests for `http-client-openssl`?

alexbiehl: @snoyberg What's the motivation for removing them?

snoyberg: The tests fail because openssl isn't being configured to perform these checks. Meanwhile, http-client-tls _is_ performing these checks, the tests pass, and would prevent against regressions in the future.

alexbiehl: Thanks for explaining. Let's remove the tests then.

snoyberg: OK, let's try with https://github.com/snoyberg/http-client/pull/423

PR #211, opened 2016-07-06
Update module comment to 'checkResponse'.

seagreen:


seagreen:
Not sure if one word fixes are worth the trouble, let me know if not.

snoyberg:
The fix is definitely appreciated! However, this points out that the entire previous sentence is in fact wrong: the default is no longer to throw exceptions, and therefore `checkResponse` doesn't even need to be mentioned. Would you be able to update the PR to just remove the now-incorrect phrase?

seagreen:
No wonder I was confused, I had the changelog open and was trying to resolve that paragraph with what I thought had changed:)

Paragraph removed; commit amended and force pushed.

snoyberg:
Thanks!

PR #222, opened 2016-08-29
http-conduit: Update haddock example for 2.2.*

bergmark:
This also removes the use of an irrefutable pattern, rethrowing other
exceptions instead.


snoyberg:
Thanks!

Issue #225, opened 2016-09-01
The closeConnection method for tls connections should not be called multiple times


cdepillabout:
In my application I make two successive connections to Facebook's Graph API using `http-conduit`. I'm experiencing a bug when making two calls one after the other.  The second call is throwing a TLS error in the middle of the conversation for some reason.  I think I've figured out why it is happening, but the explantion is pretty long, so please bear with me.

`http-conduit` is using `http-client-tls` and `http-client`.  `http-client-tls` is using the `connection` library to actually create connections.

`http-client-tls`'s [`convertConnection`](https://github.com/snoyberg/http-client/blob/517e7f3f5ad44ad8fa2775d5d7bd373faaf0de53/http-client-tls/Network/HTTP/Client/TLS.hs#L102) function is used to convert a `connection` [`Network.Connection.Types.Connection`](https://github.com/vincenthz/hs-connection/blob/3db5a4924ca23516a49699d27ef376385c0ef819/Network/Connection/Types.hs#L85) to an `http-client` `Connection`.

`http-client-tls`'s `convertConnection` function is using [`makeConnection`](https://github.com/snoyberg/http-client/blob/517e7f3f5ad44ad8fa2775d5d7bd373faaf0de53/http-client/Network/HTTP/Client/Connection.hs#L85) to create the `http-client` `Connection`.

The following is the `convertConnection` function:

```haskell
convertConnection :: NC.Connection -> IO Connection
convertConnection conn = makeConnection
    (Network.Connection.connectionGetChunk conn)
    (Network.Connection.connectionPut conn)
    -- Closing an SSL connection gracefully involves writing/reading
    -- on the socket.  But when this is called the socket might be
    -- already closed, and we get a @ResourceVanished@.
    (Network.Connection.connectionClose conn `Control.Exception.catch` \(_ :: IOException) -> return ())
```

Note how `makeConnection`'s `c` argument is becoming the following:

```haskell
-- Closing an SSL connection gracefully involves writing/reading
-- on the socket.  But when this is called the socket might be
-- already closed, and we get a @ResourceVanished@.
(Network.Connection.connectionClose conn `Control.Exception.catch` \(_ :: IOException) -> return ())
```

`makeConnection` looks like this:

```haskell
makeConnection :: IO ByteString -- ^ read
               -> (ByteString -> IO ()) -- ^ write
               -> IO () -- ^ close
               -> IO Connection
makeConnection r w c = do
    istack <- newIORef []
[...]

PR #226, opened 2016-09-02
Check for connection status in finalizer

Yuras:
It is fix for #225 
It also contains related patch to cleanup test suit, please let me know if you want me to send it separately.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -93,7 +93,12 @@ makeConnection r w c = do
     -- already closed connection.
     closedVar <- newIORef False
 
-    _ <- mkWeakIORef istack c
+    let close = do
+          closed <- atomicModifyIORef closedVar (\closed -> (True, closed))
+          unless closed $
+            c
+
+    _ <- mkWeakIORef istack close
     return $! Connection
         { connectionRead = do
             closed <- readIORef closedVar
@@ -113,11 +118,7 @@ makeConnection r w c = do
             when closed $ throwHttp ConnectionClosed
             w x
 
-        , connectionClose = do
-            closed <- readIORef closedVar
-            unless closed $
-              c
-            writeIORef closedVar True
+        , connectionClose = close
         }
 
 -- | Create a new 'Connection' from a 'Socket'.
--- a/http-client/test-nonet/Network/HTTP/ClientSpec.hs
+++ b/http-client/test-nonet/Network/HTTP/ClientSpec.hs
@@ -2,14 +2,15 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 module Network.HTTP.ClientSpec where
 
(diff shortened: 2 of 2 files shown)


Yuras:
@cdepillabout Could you please check that it fixes the issue for you.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -93,7 +93,13 @@ makeConnection r w c = do
     -- already closed connection.
     closedVar <- newIORef False
 
-    _ <- mkWeakIORef istack c
+    let close = do
+          closed <- readIORef closedVar
```
snoyberg: I'm not certain it's necessary, but to avoid a potential race condition, what about using `atomicModifyIORef` to read the closed state and immediately write back `True`?

snoyberg: I realize that's a change to how the code works now.

http-client/test-nonet/Network/HTTP/ClientSpec.hs
```diff
@@ -28,13 +36,13 @@ redirectServer inner = bracket
         (N.runTCPServer (N.serverSettingsTCPSocket lsocket) app)
         (const $ inner port)
   where
-    app ad = do
-        _ <- forkIO $ forever $ N.appRead ad
-        forever $ do
+    app ad = Async.race_
+        (forever (N.appRead ad))
+        (silentIOError $ forever $ do
```
snoyberg: What's the motivation behind this change? It's unclear why it's happening.

http-client/Network/HTTP/Client/Connection.hs
Yuras: done

http-client/test-nonet/Network/HTTP/ClientSpec.hs
Yuras: It is described briefly in the 9a39b4b87943ec5e7fdffe3292195dd8bcf51caa commit message.

Basically, `forkIO $ forever (N.appRead ad)` thread is never killed, so it performs reading from already closed socket. To trigger it, add `performGC` at the end of test suit to force cached connections to be released -- as a result write loop fails (resource vanished), `app ad` exits, server socket is closed, `appRead ad` fails with either resource vanished or bad descriptor.

In any case, I'm open for better solution.

snoyberg: Nah, that seems fine, thanks for the pointer. Let's wait for Travis.

cdepillabout:
I'm not going to be able to test this until Tuesday.  But the fix looks pretty close to what I described in #225, which did work for me, so I'd bet that this fix is good :+1.

snoyberg:
Travis failure on GHC 7.4.2

snoyberg:
Thanks!

PR #232, opened 2016-10-04
Don't hardcode version number number in the link.

psibi:
The new link always points to the latest lts and the package in it.


snoyberg:
Thanks!

PR #235, opened 2016-10-25
Add missing colon in Host header

robbinch:
I think there is a missing colon here when setting Host header while using https proxy.


robbinch:
Great! Note that stack-1.2.0 cannot use https proxy because of this.

snoyberg:
Merged. You may want to ping the Stack team (or send a PR) to get this new version of http-client in use there for the next release.

Issue #237, opened 2016-10-27
httpLbs does not throw exception for 401


MasterMastic:
The [documentation](https://haskell-lang.org/library/http-client) states "By default, every request that generates a non-2XX response will generate a runtime exception.", however:

```
ghci> import qualified Network.HTTP.Client as Http
ghci> Http.httpLbs (Http.parseRequest_ "http://httpbin.org/status/401") =<< Http.newManager Http.defaultManagerSettings
Response {responseStatus = Status {statusCode = 401, statusMessage = "UNAUTHORIZED"}, responseVersion = HTTP/1.1, responseHeaders = [("Server","nginx"),("Date","Thu, 27 Oct 2016 10:38:22 GMT"),("Content-Length","0"),("Connection","keep-alive"),("Access-Control-Allow-Origin","*"),("WWW-Authenticate","Basic realm=\"Fake Realm\""),("Access-Control-Allow-Credentials","true")], responseBody = "", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
```

snoyberg:
Good catch, this was a change in http-client 0.5. You interested in sending
a PR with an update to the docs?

MasterMastic:
So this is intentional? how come? and is there any easy way to throw an `HttpException` for it?

I wouldn't mind updating the docs, but is 401 the only exception for an error-code that doesn't throw exceptions?

mrkkrp:
Sorry for being dumb, but looking at the change log for version 0.5, I see:

* Remove instance Default Request
* Modify instance IsString Request to use parseRequest instead of parseUrlThrow
* Clean up the HttpException constructors
* Rename checkStatus to checkResponse and modify type
* Fix the ugly magic constant workaround for responseTimeout
* Remove getConnectionWrapper
* Add the HttpExceptionRequest wrapper so that all exceptions related to a request are thrown with that request's information

From all of these, only the last one seems to be somewhat relevant, but it says that a wrapper was added, even with the wrapper, it should still throw for non 2xx status codes something, but in the shown example it doesn't throw at all.

snoyberg:
It's from the change to the IsString instance. The discussion around this
was: https://github.com/snoyberg/http-client/issues/193

mrkkrp:
Ah well, I see. Would be helpful to add this change to the change log, at least retroactively, like “stopped throwing exceptions for non-2xx response status codes”.

So we need to update:

1. Changelog;
2. Docs of the package itself;
[...]

PR #238, opened 2016-10-29
Improve docs and fix deprecation warnings

mrkkrp:
Should hopefully help avoid confusion expressed in #237.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -86,8 +86,9 @@ with this language extension enabled.
 
 There are a few important caveats to mention about this library:
 
-* By default, any non-2XX status code response results in a runtime exception.
-  See the examples of `setRequestIgnoreStatus` below for more information
+* By default, any non-2XX status code response won't result in a runtime
+  exception contrary to previous behaviour of the library (before version
+  0.5).
 * By default, http-client will respect the `http_proxy` and `https_proxy`
   environment variables. See the proxy examples below for information on how to
   bypass this.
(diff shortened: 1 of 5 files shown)


TUTORIAL.md
```diff
@@ -86,8 +86,12 @@ with this language extension enabled.
 
 There are a few important caveats to mention about this library:
 
-* By default, any non-2XX status code response results in a runtime exception.
-  See the examples of `setRequestIgnoreStatus` below for more information
+* By default, any non-2XX status code response results in a runtime
```
snoyberg: This is obviously a good change, but I'd like to suggest making it slightly better: change this to say that by default status code will not result in runtime exceptions, contrary to previous behavior of the library.

TUTORIAL.md
```diff
@@ -311,7 +315,7 @@ main = do
 
 ## Non-2XX responses
 
-By default, every request that generates a non-2XX response will generate a
+By default, every request that generates a non-2XX response will throw a
```
snoyberg: Similarly here, let's just fix the offending content instead of correcting it later in the document.

TUTORIAL.md
```diff
@@ -507,18 +531,18 @@ main :: IO ()
 main = do
     manager <- newManager tlsManagerSettings
 
-    request <- parseUrl "http://httpbin.org/get"
+    request <- parseRequest_ "http://httpbin.org/get"
```
snoyberg: I think you need either `let request = parseRequest_` or `request <- parseRequest`.

snoyberg: Similar comments apply below

mrkkrp:
I made the requested changes.

snoyberg:
Thanks!

PR #244, opened 2016-11-22
Use redirectCount set through managerModifyRequest

### LINKED ISSUE #243: Setting redirectCount to 0 with managerModifyRequest does not work
I am using stackage's lts-7.3 which pulls http-client 0.4.31.1 by default and setting redirectCount to 0 does not work as expected: Redirection is still followed. I can see this behaviour has been fixed by https://github.com/snoyberg/http-client/commit/11a9a6273375e67c157b4b0a8a359c73029512cb Would it be possible to backport it so that it is available using lts-7.X dependency set?

abailly:
fixes #243 for maintenance branch.

### DIFF
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE DeriveFunctor #-}
-{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DeriveTraversable  #-}
 {-# LANGUAGE DeriveFoldable #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE DeriveFunctor #-}
-{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DeriveTraversable  #-}
```
snoyberg: Unintended change?

http-client/test/Network/HTTP/ClientSpec.hs
```diff
@@ -1,12 +1,12 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Network.HTTP.ClientSpec where
 
-import           Control.Exception         (toException)
-import           Network                   (withSocketsDo)
-import           Network.HTTP.Client
-import           Network.HTTP.Types        (status200, status405)
-import           Test.Hspec
-import           Data.ByteString.Lazy.Char8 () -- orphan instance
```
snoyberg: Were these changes needed?

http-client/Network/HTTP/Client.hs
abailly: Yeah, sorry for that. I had my stylish-haskell on, it was late, I tried to fix it manually... I will do another PR to remove that.

snoyberg:
Not a big deal tbh.

abailly:
I know, but I also know it adds unwanted noise down the line :)

Issue #248, opened 2016-12-14
http-client should allow to log requests and responses


loewenstein:
I would like to be able to see all requests and responses in a debug log (e.g. on STDERR or STDOUT).

I have seen https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Types.hs#L672 and https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Types.hs#L497. However running the BodyReader and printing the result consumes the body, i.e. it is no longer available for the issuer of the request.

If the two hooks are actually the right ones to use (I doubt that, it looked like a workaround in the beginning), we should be able to modify the Response (just as we can modify the request). That way one could feed the body back into a new otherwise identical Response.
This would have to be solved even if we provide separate hooks though.

What do you think?

With some amount of guidance I might be able to come up with a PR.

Edit: Maybe I should add, I am using http-client through servant-client. Hence my ability to just call a different request execution method is rather limited.

snoyberg:
These definitely weren't intended for the logging use case, but if it fits, I'd probably say to stick with that. If you really need to log the response bodies, I'd be open to an extra method in ManagerSettings for post-processing a `Response` value.

loewenstein:
Well, it doesn't work because the body gets consumed from me printing it to stderr. At least that is my interpretation of what I see.

"post-processing a `Response` value" would be a separate method like `managerModifyRequest`, i.e. `managerModifyResponse :: ManagerSettings -> Response -> IO Response`?

So logging would go into the two separate modify methods. Sounds good to me.
If you have a pointer to where these get called should get called I would be happy to propose a PR.

I guess the logging functionality would have to consume the body and provide a Response with an in memory version of the body, right? I guess I should limit body logging to a limited `Content-Length` then.

snoyberg:
> Well, it doesn't work because the body gets consumed from me printing it to stderr. At least that is my interpretation of what I see.

I agree. That's what I was getting at with "if you need to log the response bodies." Many use cases would simply want to log the response metadata (status/headers), in which case the current API is sufficient. But your use case may not be that.

Yes, I think `managerModifyResponse` is the right thing. And it sounds like a PR would only need to add that one method, and it could be called from the same place as `checkResponse`.

You're correct about the in memory representation, though an alternative would be streaming to a temporary file. But to be clear, I think the logging functionality itself would _not_ end up in http-client, correct?

loewenstein:
[...]

PR #249, opened 2016-12-16
Modify response

loewenstein:
Add a `modifyResponse` hook similar to the one for `Request` modification.

It expects a function of type `Response BodyReader -> IO (Response BodyReader)` that will be executed just after `checkResponse` was called.

This will allow to implement #248 outside of the http-client library.

Unfortunately, and that is why I still put a WiP in front, my test for body modification does not run through. @snoyberg Any idea what I am doing wrong?

### DIFF
--- a/http-client/Network/HTTP/Client/Body.hs
+++ b/http-client/Network/HTTP/Client/Body.hs
@@ -7,6 +7,7 @@ module Network.HTTP.Client.Body
     , makeUnlimitedReader
     , brConsume
     , brEmpty
+    , constBodyReader
     , brAddCleanup
     , brReadSome
     , brRead
@@ -51,6 +52,14 @@ brReadSome brRead' =
 
 brEmpty :: BodyReader
 brEmpty = return S.empty
+
+constBodyReader :: [S.ByteString] -> IO BodyReader
+constBodyReader input = do
+  iinput <- newIORef input
+  return $ atomicModifyIORef iinput $ \input' ->
+        case input' of
+            [] -> ([], S.empty)
+            x:xs -> (xs, x)
 
 brAddCleanup :: IO () -> BodyReader -> BodyReader
 brAddCleanup cleanup brRead' = do
--- a/http-client/test/Network/HTTP/ClientSpec.hs
+++ b/http-client/test/Network/HTTP/ClientSpec.hs
@@ -2,8 +2,10 @@
 module Network.HTTP.ClientSpec where
 
 import           Network                   (withSocketsDo)
+import qualified Data.ByteString.Char8        as BS
 import           Network.HTTP.Client
 import           Network.HTTP.Types        (status200, found302, status405)
+import           Network.HTTP.Types.Status
 import           Test.Hspec
 import           Control.Applicative       ((<$>))
 import           Data.ByteString.Lazy.Char8 () -- orphan instance
@@ -45,6 +47,20 @@ spec = describe "Client" $ do
             man <- newManager defaultManagerSettings
             res <- httpLbs req man
             responseStatus res `shouldBe` found302
+
+    context "managerModifyResponse" $ do
+      it "allows to modify the response status code" $ do
+        let modify :: Response BodyReader -> IO (Response BodyReader)
+            modify res = do
+              return res {
+                responseStatus = (responseStatus res) {
+                  statusCode = 201
+                }
+              }
+            settings = defaultManagerSettings { managerModifyResponse = modify }
+        man <- newManager settings
+        res <- httpLbs "http://httpbin.org" man
+        (statusCode.responseStatus) res `shouldBe` 201
 
     context "managerModifyRequest" $ do
         it "port" $ do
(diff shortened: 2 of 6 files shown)


snoyberg:
Can you clarify what error message you're seeing?

loewenstein:
Well, first of all the test hangs for quite some time.

In the end the test does not appear in the output, neither in red nor black or any color:
```
managerModifyResponse
      allows to modify the response status code
```

The whole suite fails with:
```
Test suite failure for package http-client-0.5.4
    spec:  exited with: ExitFailure (-9)
Logs printed to console
```

Full output:
```
13:24 $ stack test http-client
http-client-0.5.4: test (suite: spec)

Network.HTTP.Client
  Client
    works
    method in URL
      success
      failure
    redirects
      follows redirects
      allows to disable redirect following
    managerModifyResponse
      allows to modify the response status code

http-client-0.5.4: test (suite: spec-nonet)


Network.HTTP.Client.Body
  BodySpec
    chunked, single
    chunked, pieces
    chunked, raw
    chunked, pieces, raw
    length, single
    length, pieces
    gzip
Network.HTTP.Client.Cookie
  CookieSpec
    cookie equality - case insensitive Eq
    domainMatches - case insensitive
    domainMatches - case insensitive, partial
Network.HTTP.Client.Headers
  HeadersSpec
    simple response
    Expect: 100-continue (success)
    Expect: 100-continue (failure)
    100 Continue without expectation is OK
Network.HTTP.Client.RequestBody
  streamFile
    works
Network.HTTP.Client.Request
  case insensitive scheme
    http://example.com
    httP://example.com
    HttP://example.com
    HttPs://example.com
    ftp://example.com
  authentication in url
    passes validation
    add username/password to headers section
  applyBasicProxyAuth
    Should add a proxy-authorization header
    Should add a proxy-authorization header with the specified username and password.
  extract credentials from a URI
    fetches non-empty username before the first ':'
    after ':' is considered password
    decodes username special characters per RFC3986
    decodes password special characters per RFC3986
    no auth is empty
  requestBuilder
    sends the full request, combining headers and body in the non-streaming case
    sends only headers and returns an action for the body on 'Expect: 100-continue'
Network.HTTP.Client.Response
  ResponseSpec
    basic
    no length
    chunked
    gzip
Network.HTTP.Client
  Client
    fails on empty hostnames #40
      http://
      https://
      http://:8000
      https://:8001
    redirecting #41
    redirectCount=0
    connecting to missing server gives nice error message
    extra headers after 100 #49
      False
      True
    early close on a 413
    length zero and chunking zero #108
    length zero and chunking
    length and chunking
    withResponseHistory and redirect
    should not write to closed connection

Finished in 2.6981 seconds
50 examples, 0 failures

Test suite failure for package http-client-0.5.4
    spec:  exited with: ExitFailure (-9)
Logs printed to console
```

http-client/Network/HTTP/Client/Body.hs
```diff
@@ -52,6 +53,15 @@ brReadSome brRead' =
 brEmpty :: BodyReader
 brEmpty = return S.empty
 
+constBodyReader :: [S.ByteString] -> IO BodyReader
+constBodyReader input = do
+  iinput <- newIORef input
+  ioutput <- newIORef []
```
snoyberg: What's the purpose of `ioutput`? It appears to never be used.

http-client/test/Network/HTTP/ClientSpec.hs
```diff
@@ -46,6 +49,33 @@ spec = describe "Client" $ do
...
+        man <- newManager settings
+        res <- httpLbs "http://httpbin.org" man
+        (statusCode.responseStatus) res `shouldBe` 201
+
+      it "modifies the response body" $ do
+        let modify :: Response BodyReader -> IO (Response BodyReader)
+            modify res = do
+              return res {
+                responseBody = do
+                  reader <- constBodyReader [BS.pack "modified response body"]
```
snoyberg: This is the bug. You've defined the `responseBody` to be an action which, every time it runs, it creates a new `BodyReader` and returns exactly one chunk. You need to move the `constBodyReader` call to outside the `responseBody` definition.

snoyberg:
I've reviewed the code, and left some comments which should explain the looping issue.

http-client/Network/HTTP/Client/Body.hs
loewenstein: You are right. It is never used. I started from a copy of dummyConnection ioutput was accidentally left in there.

http-client/test/Network/HTTP/ClientSpec.hs
loewenstein: Oh. Makes sense, now that you pointed me at it.

Out of curiosity. Who is continuously accessing `responseBody` in the test case?

loewenstein:
@snoyberg Just pushed the fixes. Travis has a problem with [GHC 7.6.3 C](https://travis-ci.org/snoyberg/http-client/jobs/185124620). However I don't see a relation to my changes.

```
cabal: Error: some packages failed to install:
criterion-1.1.4.0 depends on js-jquery-3.1.1 which failed to install.
js-jquery-3.1.1 failed during the configure step. The exception was:
user error (The package requires Cabal library version -any && >=1.18 but no
suitable version is installed.)
```

From my point of view I am done. I will remove WiP prefix from PR title.

loewenstein:
Thanks for the quick review and help fixing the problem.

snoyberg:
It's OK to ignore that 7.6.3 error.

PR #250, opened 2016-12-22
Document escape rules.

alexanderkjeldaas:


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -110,6 +110,8 @@ parseUrlThrow =
 --
 -- 'Request' created by this function won't cause exceptions on non-2XX
 -- response status codes.
+-- 
+-- The URI string will be escaped, and should not be URI escaped.
```
snoyberg: This is definitely an improvement, thank you! However, what do you think about including a small example of how this works in practice?

PR #254, opened 2017-01-02
STM-based manager

snoyberg:
This code is hopefully much easier to review, and more obviously correct via decoupling the pool implementation from the surrounding code. Not for immediate merge, but definitely interested in review from others.

### DIFF
--- a/http-client/Data/KeyedPool.hs
+++ b/http-client/Data/KeyedPool.hs
@@ -107,8 +107,9 @@ createKeyedPool
     -> (resource -> IO ()) -- ^ destroy a resource
     -> Int -- ^ number of resources per key to allow in the pool
     -> Int -- ^ number of resources to allow in the pool across all keys
+    -> (SomeException -> IO ()) -- ^ what to do if the reaper throws an exception
     -> IO (KeyedPool key resource)
-createKeyedPool create destroy maxPerKey maxTotal = do
+createKeyedPool create destroy maxPerKey maxTotal onReaperException = do
     var <- newTVarIO $ PoolOpen 0 Map.empty
 
     -- We use a different IORef for the weak ref instead of the var
@@ -126,7 +127,7 @@ createKeyedPool create destroy maxPerKey maxTotal = do
     -- cleanup will be triggered.
 
     -- Ensure that we have a normal masking state in the new thread.
-    _ <- forkIOWithUnmask $ \restore -> restore $ reap destroy var
+    _ <- forkIOWithUnmask $ \restore -> keepRunning $ restore $ reap destroy var
     return KeyedPool
         { kpCreate = create
         , kpDestroy = destroy
@@ -135,6 +136,11 @@ createKeyedPool create destroy maxPerKey maxTotal = do
         , kpVar = var
         , kpAlive = alive
         }
+  where
+    keepRunning action =
+        loop
+      where
+        loop = action `catch` \e -> onReaperException e >> loop
 
 -- | Make a 'KeyedPool' inactive and destroy all idle resources.
 destroyKeyedPool' :: (resource -> IO ())
--- a/http-client/Data/KeyedPool.hs
+++ b/http-client/Data/KeyedPool.hs
@@ -104,7 +104,10 @@ plistFromList xs =
 createKeyedPool
     :: Ord key
     => (key -> IO resource) -- ^ create a new resource
-    -> (resource -> IO ()) -- ^ destroy a resource
+    -> (resource -> IO ())
+       -- ^ Destroy a resource. Note that exceptions thrown by this will be
+       -- silently discarded. If you want reporting, please install an
+       -- exception handler yourself.
     -> Int -- ^ number of resources per key to allow in the pool
     -> Int -- ^ number of resources to allow in the pool across all keys
     -> (SomeException -> IO ()) -- ^ what to do if the reaper throws an exception
@@ -148,7 +151,7 @@ destroyKeyedPool' :: (resource -> IO ())
                   -> IO ()
 destroyKeyedPool' destroy var = do
     m <- atomically $ swapTVar var PoolClosed
-    F.mapM_ destroy m
+    F.mapM_ (ignoreExceptions . destroy) m
 
 -- | Run a reaper thread, which will destroy old resources. It will
 -- stop running once our pool switches to PoolClosed, which is handled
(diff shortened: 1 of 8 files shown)


http-client/Data/KeyedPool.hs
```diff
@@ -0,0 +1,300 @@
...
+createKeyedPool
+    :: Ord key
+    => (key -> IO resource) -- ^ create a new resource
+    -> (resource -> IO ()) -- ^ destroy a resource
+    -> Int -- ^ number of resources per key to allow in the pool
+    -> Int -- ^ number of resources to allow in the pool across all keys
+    -> IO (KeyedPool key resource)
+createKeyedPool create destroy maxPerKey maxTotal = do
+    var <- newTVarIO $ PoolOpen 0 Map.empty
+    _ <- forkIO $ reap destroy var
```
Yuras: Should `forkIO` be protected with `bracketOnException`? Looks like `createKeyedPool` is called without async exceptions masked, so async exception may occur between `forkIO` and `mkWeekIORef`. Also, is in desired to inherit the current thread mask? I'd use `forkIOWithUnmask` for sanity.

snoyberg: Good catches, thanks!

Yuras:
Let me use the chance to rise a concern regarding the reaper thread. Feel free to ignore it if you thing it is out of scope.

Basically, forking an utility thread in a library is not a good patten because the thread may fail silently (the notice will appear in stdout, but still). You are careful to ensure an exception in resource destroying action won't break reaper, but sooner or later something will go wrong. When (if) [we'll have](https://phabricator.haskell.org/D2790) working `HeapOverflow` exception, it will become impossible to ensure utility thread never fails. Better approach could be to provide an option to run reaper manually. What do you think? Or is it too paranoid?

http-client/Data/KeyedPool.hs
snoyberg: I just rearranged the code instead of `bracketOnException`, with a comment justifying why, I think this is correct.

snoyberg:
I think it's a fair concern to raise, but it's also something manageable. What about just installing an exception handler which will catch anything that goes wrong in the thread, optionally log it somewhere (maybe based on some passed-in setting), and then restart the reaper?

Yuras:
Yes, restarting the reaper seems to be a good solution.

snoyberg:
Awesome, how's that?

Yuras:
Looks good, thank you!

snoyberg:
No, thank you for the review!

agrafix:
It would be really nice to add some unit tests to `Data.KeyedPool` ensuring that the various assertions currently made in comments actually hold, to test exception safety and that resources are actually cleaned up.

http-client/Data/KeyedPool.hs
```diff
@@ -0,0 +1,316 @@
...
+                toDestroy' = toDestroy . (map snd stale++)
+                toKeep' =
+                    case plistFromList notStale of
+                        Nothing -> toKeep
+                        Just x -> toKeep . ((key, x):)
+        let (toKeep, toDestroy) = findStale' id id (Map.toList m)
+        let idleCount' = idleCount - length toDestroy
+        return (PoolOpen idleCount' toKeep, toDestroy)
+
+    safeDestroy x = destroy x `catch` \(_ :: SomeException) -> return ()
```
agrafix: I don't know if we actually want to silently catch all exceptions here. We should at least somehow log them because it's very important to know if destroying a resource didn't work for some reason.

snoyberg: Here are the options I see:

* Print a message when an exception is caught. I think this is bad behavior in general: libraries shouldn't pollute standard output or error like that.
* Take in an extra parameter of "what to do if the destroy action throws an exception." But frankly, there's no need to do that: the destroy action itself is already a parameter, and a user of the library can trivially add the `catch` him/herself. Maybe there's an argument for forcing the user to think about this though.

http-client/Data/KeyedPool.hs
```diff
@@ -0,0 +1,316 @@
...
+
+-- | Check out a value from the 'KeyedPool' with the given key.
+--
+-- This function will internally call 'mask_' to ensure async safety,
+-- and will return a value which uses weak references to ensure that
+-- the value is cleaned up. However, if you want to ensure timely
+-- resource cleanup, you should bracket this operation together with
+-- 'managedRelease'.
+takeKeyedPool :: Ord key => KeyedPool key resource -> key -> IO (Managed resource)
+takeKeyedPool kp key = mask_ $ join $ atomically $ do
```
Yuras: Why `mask_` here? Exception safety of this function relies on weak finalizer, so I think it is enough to mask async exceptions from `kpCreate` to `mkWeakIORef` in the `do` block below. Not a big deal though.

http-client/Data/KeyedPool.hs
agrafix: You are right, there not really good options out of this. Maybe a note in the documentation about the exception behavior would be nice.

agrafix:
@Yuras : I've observed this, but we tracked it down to the problem in the `connection` package. TLS-Handshake failures are highly dependent on many situations, so maybe I was just looking at a time where they where very common...

http-client/Data/KeyedPool.hs
```diff
@@ -0,0 +1,316 @@
...
+        loop
+      where
+        loop = action `catch` \e -> onReaperException e >> loop
+
+-- | Make a 'KeyedPool' inactive and destroy all idle resources.
+destroyKeyedPool' :: (resource -> IO ())
+                  -> TVar (PoolMap key resource)
+                  -> IO ()
+destroyKeyedPool' destroy var = do
+    m <- atomically $ swapTVar var PoolClosed
```
snoyberg: It's part of the core contract of STM that anything inside an atomically block either completely succeeds or completely fails. I'd overall be -1 on a comment here, it's integral to the understanding of STM, and the rest of the module definitely requires understanding of how STM works.

http-client/Data/KeyedPool.hs
```diff
@@ -0,0 +1,316 @@
...
+    F.mapM_ destroy m
```
snoyberg: I don't know about the weak finalizer point, but we definitely need to catch exceptions here just like we do above.

http-client/Data/KeyedPool.hs
snoyberg: Without the `mask_`, an exception could be thrown between the completion of the `atomically` call and the completion of the `join` call, thereby preventing the `mkWeakIORef` call from ever being made.

http-client/Data/KeyedPool.hs
Yuras: I mean that the STM action may (?) block, so `atomically` could be interrupted by async exception. Then the following line will not be called at all.

Yuras: Oops, it is used only as a finalizer, so no async  exception here. Sorry, please ignore the previous comment.

PR #255, opened 2017-01-05
Better get addr info exception

snoyberg:


neongreen:
This is nice! Anything that prevents it from being merged? (I haven't looked at whether the merge conflict is serious or not)

snoyberg:
I think I shelved it due to concerns of breaking backwards compatibility in exceptions (something very hard to test against), and hopes that the network library itself would provide this functionality in the future. I'm not opposed to seeing it move forward though. Do you have a specific motivation for seeing this happen in mind?

neongreen:
Nope, just tired of getAddrInfo errors and hoped that someone would do the job for me :smiling_imp: On the other hand, I just noticed that the `network` library people want to move forward with `getAddrInfo` error messages (https://github.com/haskell/network/issues/332) so maybe this PR is not needed after all.

neongreen:
Even more interesting: the fix is actually included in network-2.7, so this PR is not needed at all and can be close.

snoyberg:
Cool. I'll close this.

PR #259, opened 2017-02-03
Lenses

reactormonk:
Not my code, I just found it lying around, might be useful.

### DIFF
--- a/http-client/Network/HTTP/Client/Lens.hs
+++ b/http-client/Network/HTTP/Client/Lens.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE RankNTypes #-}
+
 -- | Optics for @http-client@ types
 module Network.HTTP.Client.Lens
   ( -- * 'Request' lenses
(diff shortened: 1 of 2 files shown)


snoyberg:
lens is too heavy a dependency for this package. Making a separate package would make more sense.

reactormonk:
Where did you see a dependency on lens?

reactormonk:
I'll see if I can remove the dependency, shouldn't be too hard.

reactormonk:
Done, no more Lens dependency.

snoyberg:
I'm guessing it's not updating because I closed the PR, so I'll open it for now. That said: I'm not too excited about this move here. It's an arbitrary cut-off point between which lenses are and are not included based on presence of prisms, and it's not clear how to best use this module due to naming conflicts with the main module. I think overall I'd rather see this develop into something useful as a separate package.

reactormonk:
Naming conflicts can be solved, I've seen an `L` postfix. I've removed the exception package mainly because from what I understand, the exception handling has been changed. It would be possible to add them in again, but I figured the other lenses are more useful and should be included first, and the rest later.

Prisms can be defined [by depending on profunctors](https://github.com/ekmett/lens/wiki/How-can-I-write-lenses-without-depending-on-lens%3F#prism). It's also possible to lift lenses into prisms with the appropriate library, which would require some additional grunt work to use them. But allows you to write them without having to depend on `profunctors`.

neongreen:
> It's also possible to lift lenses into prisms with the appropriate library

I'm curious, what library?

snoyberg:
This isn't really changing my overall opinion: there are still design
points to be worked out, there are potentially new dependencies to be
added, the API may not be stable, and there's no inherent benefit to
including the module in this package instead of a new package.

reactormonk:
@neongreen From what I understand, https://hackage.haskell.org/package/lens-4.15.1/docs/Control-Lens-Prism.html#v:prism should work.

neongreen:
@reactormonk ah, you meant `prism`. Okay.

Issue #153, opened 2015-09-08
support socks5 proxy specification via http_proxy variable


ibukanov:
Curl and other libraries allow to specify socks5 proxy using [http_proxy variable](http://curl.haxx.se/docs/manpage.html#PROXY), like in:

```
socks5://host:port
socks5h://host:port
```

where the latter form forces DNS resolution over SOCKS protocol. It would be nice for http-client to support this. This way I could use programs compiled with http-client together with `ssh -D`.

snoyberg:
I have no objection to such a feature, and it should be relatively easy to add to the existing environment variable parsing code. Want to take a crack at a PR?

ibukanov:
My Haskell experience is too newbie for PT . In fact I was hit by that when tried to use Elm tools behind a firewall...

RobertFischer:
@snoyberg -- I'm happy to take a swing at this. Can you give me a quick pointer so that I am moving in the right direction?

RobertFischer:
(I believe this is the underlying issue for Stack not accepting my sock5 proxy, which is why I am motivated to fix it.)

snoyberg:
Check out envHelper in https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Manager.hs

RobertFischer:
PR submitted: https://github.com/snoyberg/http-client/pull/262

snoyberg:
PR merged, thanks!

kirelagin:
For any future readers who get as confused as I was: the tldr; is that after a couple of other changes the current situation is that `socks5://` in `http_proxy` works, but only with `newTlsManager` from `http-client-tls` (and _not_ with the simple manager from `http-client`).

PR #262, opened 2017-02-12
Issue #153: Added socks5 and socks5h support; tests pass

RobertFischer:
I didn't see exactly how to test the usage of a SOCKS5 proxy via the environment variable -- everything that I saw said that it just should work. If you point me in the direction of how to write a test, I'm happy to do so.

### DIFF
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -509,6 +509,8 @@ envHelper name eh = do
             (p, muserpass) <- maybe invalid return $ do
                 uri <- case U.parseURI str of
                     Just u | U.uriScheme u == "http:" -> return u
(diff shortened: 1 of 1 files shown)


snoyberg:
I don't understand the comment, can you clarify what you mean?

RobertFischer:
From what I'm understanding, the change that I made will accept "socks5://" and "socks5h://" as the prefix to the `HTTPS_PROXY` and `HTTP_PROXY` strings. But I didn't see how to write a test to prove that using those URL prefixes would actually connect us to a SOCKS5 proxy and successfully interact with it.

snoyberg:
It looks fine to me. Ideally we would have some kind of a docker-compose based test that actually sets up proxy servers, but that doesn't exist right now.

Issue #268, opened 2017-03-20
Support for Windows proxy settings


varosi:
Currently proxy settings are get from HTTPS_PROXY and HTTP_PROXY env. variables which are not a standard in Windows OS. Every tool created with http-client under Windows is suffering from this problem. For every tool the developer have to ask users of the tool to set those env. variables if they have proxies which is inappropriate under Windows.
Standard places to set proxies there are under LAN settings accessible by any browser under Windows. 
Here are some info about how it could be extracted:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa384095(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/aa384096(VS.85).aspx

varosi:
Second method return proxy info for sure, at least on my Windows 8.

varosi:
This could help, too
https://hackage.haskell.org/package/HTTP-4000.3.6/docs/Network-HTTP-Proxy.html#t:Proxy

it's returning correct value on my work station.

snoyberg:
Would you be interested in sending a PR for this functionality?

varosi:
Yes, I will try to next weeks.

varosi:
Is it Okay for you to bring http://hackage.haskell.org/package/HTTP dependency to http-client, because of Network.HTTP.Proxy ?

snoyberg:
No, we can't add that dep

varosi:
Other way is to copy paste proxy reading from there, which is not beautiful.

varosi:
Okay, I'll create a new library that is dealing with finding proxies, so this functionality could be reused.

snoyberg:
I'd really rather the functionality go straight into http-client, I already
get people complaining regularly about any added dependencies. In principle
I agree with doing things as separate libraries, I'm just trying to find a
balance that makes the least number of people upset.

snoyberg:
Addressed by #274.

PR #274, opened 2017-05-06
Support for Windows system proxy settings

varosi:
Related to https://github.com/snoyberg/http-client/issues/268

Code is copy pasted from HTTP module, have some additional fixes and integrated with http-client modules. I'm open to any suggestions for fixes.

### DIFF
--- a/http-client/Network/HTTP/Proxy.hs
+++ b/http-client/Network/HTTP/Proxy.hs
@@ -203,6 +203,7 @@ parseWindowsProxy proto s =
 -- Extract proxy settings from Windows registry. This is a standard way in Windows OS.
 systemProxy :: ProxyProtocol -> IO (HostAddress -> Maybe ProxySettings)
 systemProxy proto = do
+    -- TODO Support for proxy skipping for specific addresses, kept in "ProxyOverride" registry key
     proxy <- fetchProxy proto
     return $ const proxy
 
--- a/http-client/Network/HTTP/Proxy.hs
+++ b/http-client/Network/HTTP/Proxy.hs
@@ -143,7 +143,7 @@ systemProxyHelper envOveride prot eh = do
             toRequest Nothing                            = noEnvProxy req
             toRequest (Just (ProxySettings p muserpass)) = maybe id (uncurry applyBasicProxyAuth) muserpass
                                         req { proxy = Just p }
-    pure result
+    return result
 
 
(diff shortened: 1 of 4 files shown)


snoyberg:
I'd prefer if the other code changes be in a separate PR. Some of them are things I would rather not have included, and in any event it makes it more difficult to see what changes are introduced for this feature.

In addition, usage of `pure` seems to have broken the build for GHC < 7.10.

varosi:
What changes do you mean by "other"? What you would like not to be included?

snoyberg:
Anything not directly related to the change at hand, such as changing whitespace or reformatting imports. Some of the changes (like aligning records) is something I wouldn't include, as it makes diffs more noisy in the future.

varosi:
Okay, I'll revert reformatting, no problem.

varosi:
Done. Something else?

snoyberg:
Cursory review looks good, thanks. I'll review in more detail and hopefully merge tomorrow.

varosi:
Great! If some other changes are needed, I'll apply.

http-client/Network/HTTP/Proxy.hs
```diff
@@ -0,0 +1,355 @@
...
+systemProxyHelper :: Maybe T.Text -> ProxyProtocol -> EnvHelper -> IO (Request -> Request)
+systemProxyHelper envOveride prot eh = do
+    let envName' Nothing     = envName prot
+        envName' (Just name) = name
+
+    modifier <- envHelper (envName' envOveride)
+
+-- Under Windows try first env. variables override then Windows proxy settings
+#if defined(mingw32_HOST_OS)
+    modifier' <- systemProxy prot
```
snoyberg: It looks like it would be more efficient to check the env vars first and, if that is `Nothing`, _then_ call `systemProxy`. Right now, if `modifier` is `Just`, we'll have an unnecessary `systemProxy` call.

snoyberg:
Thanks!

http-client/Network/HTTP/Proxy.hs
varosi: hm, it's not so simple. I have two actions:
```
envProxy' :: IO (HostAddress -> Maybe ProxySettings)
systemProxy' :: IO (HostAddress -> Maybe ProxySettings)

```

How to join the two with the second depending on the result of the inner function of the first?

PR #275, opened 2017-05-10
HLint and stylish-haskell suggestions

varosi:
Those suggestions were extracted from https://github.com/snoyberg/http-client/pull/274 review.

### DIFF
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -1,9 +1,9 @@
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE RankNTypes #-}
 module Network.HTTP.Client.Manager
     ( ManagerSettings (..)
     , newManager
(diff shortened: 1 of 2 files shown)


http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -94,15 +99,15 @@ defaultManagerSettings = ManagerSettings
         case fromException e of
             Just (_ :: IOException) -> True
             _ ->
-                case fmap unHttpExceptionContentWrapper $ fromException e of
+                case unHttpExceptionContentWrapper <$> fromException e of
```
snoyberg: It looks like this change will introduce warnings in newer GHCs.

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -94,15 +99,15 @@ defaultManagerSettings = ManagerSettings
...
                     -- Note: Some servers will timeout connections by accepting
                     -- the incoming packets for the new request, but closing
                     -- the connection as soon as we try to read. To make sure
                     -- we open a new connection under these circumstances, we
                     -- check for the NoResponseDataReceived exception.
                     Just NoResponseDataReceived -> True
-                    Just IncompleteHeaders -> True
-                    _ -> False
+                    Just IncompleteHeaders      -> True
```
snoyberg: I'm -1 on such alignment changes; it encourages noisy diffs for future changes.

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -212,9 +217,7 @@ reap baton wmapRef =
     loop = do
         threadDelay (5 * 1000 * 1000)
         mmapRef <- deRefWeak wmapRef
-        case mmapRef of
```
snoyberg: I believe this form is less clear, and in particular the comment explaining what the `Nothing` case means is lost.

http-client/Network/HTTP/Client/Manager.hs
snoyberg: Note that this applies to many other diffs in this PR.

varosi:
I think that all the changes are not much useful, so I'm closing it.

PR #279, opened 2017-05-22
Allow creating a TLS-based Manager using specified settings

### LINKED ISSUE #278: newTlsManager with provided settings
It would be nice to have a variant of `newTlsManager` that uses a provided `ManagerSettings` as a base rather than the default one so that things like timeouts can be configured.

I can provide a PR if you prefer.

ivan-m:
For example, to allow for a different timeout.

Closes #278

### DIFF
--- a/http-client-tls/ChangeLog.md
+++ b/http-client-tls/ChangeLog.md
@@ -1,3 +1,10 @@
+## 0.3.5
+
+* Add `newTlsManagerWith`
+  [#278](https://github.com/snoyberg/http-client/issues/278), which
+  provides a variant of `newTlsManager` that takes a `ManagerSettings`
+  to base its settings off of.
+
 ## 0.3.4.2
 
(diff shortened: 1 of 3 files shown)


snoyberg:
Can you do a minor version bump, update the ChangeLog, and add a `@since` comment?

ivan-m:
Sure. Is there any other functions I should similarly generalise whilst I'm at it?

snoyberg:
None that I can think of.

ivan-m:
Done

ivan-m:
Is it possible for a release to be made with this please?

snoyberg:
Thanks for the ping, done.

ivan-m:
By "done", do you mean up on Hackage/Stackage? Because I'm still not seeing it there...

snoyberg:
My mistake, I'd uploaded http-client instead of http-client-tls. Should be
good now.

PR #280, opened 2017-06-23
Clarify docs regarding throwing on non-2XX requests

jhartikainen:
The docs for Network.HTTP.Simple were vague regarding what happens with non-2XX requests. 

Both `parseRequest` and `setRequestIgnoreStatus` allow you to *disable* throwing on non-2XX, which suggests the default behavior is to throw. The `IsString` didn't specify what function it was using behind the scenes, suggesting that you could create a throwing request using it (which was not the case as I found out later)

This PR adds clarification on the `IsString` behavior being the same as `parseRequest`, and also adds a note into `parseRequest` pointing at the correct function to use.


snoyberg:
Thanks!

Issue #111, opened 2015-03-18
Test suite fails on Windows


DanBurton:
As noted in #109. test-nonet/Network/HTTP/ClientSpec.hs has some failures.

ndmitchell:
appveyor may be a solution to get a proper buildbot on Windows.

hvr:
I was just pointed to this 3 year old ticket... maybe the appveyor script from #329 can help resolve this

snoyberg:
I did a little more research on this today, specifically the failing 413 test. It looks like what's happening is this:

1. Grab a listening socket
2. Fork a thread to work as a server
3. Client opens a connection to the server, sending an infinitely sized chunked request body
4. Server reads the headers, then sends a 413 (too large) response and closes the connection
5. The client tries to send more of its infinite request body, gets an exception, and moves on

Here's where we get Windows-specific failures. On Linux and OS X, the client then tries to read the response body, succeeds, and the test passes. On Windows, the client tries to read from the socket and gets an exception. I don't know why that would be the case. I've tried playing with tweaking step (4) to use the `shutdown` command instead, to no avail.

I'm tempted to simply comment out these tests on Windows. If my analysis is correct, this is exclusively a problem on the server side (closing the connection prevents further reads on the client from succeeding, which shouldn't be the case when there is data already buffered). But perhaps someone with more Windows knowledge can explain things to me.

snoyberg:
While I'd rather actually solve the failures in the test harness, if we haven't figured it out in the past few years, we aren't likely to do so going forward. The tests are now enabled for AppVeyor, so the primary issue is resolved. Closing.

PR #283, opened 2017-07-06
Remove withSocketsDo from documentation and testing

ndmitchell:
Since #107 `withSocketsDo` has been unnecessary, but it's still used in the test suite, and still documented. This PR fixes that.

I tested things, and confirmed they got no worse on Windows (0 http-conduit failures, 5 http-client failures), although the test suite is broken there (see #111) so it isn't fully checked, but it should be fine.


snoyberg:
Thanks!

PR #298, opened 2017-09-07
Support proxy for openssl

junjihashimoto:
Hi,

This PR supports http-proxy for openssl.
The test-flag(test-proxy of cabal)  is off by default.

PR #299, opened 2017-09-15
Fix a proxy-connection-bug for openssl

junjihashimoto:
Hi,

https://github.com/snoyberg/http-client/pull/298 sometimes fails,
because [this connection](https://github.com/snoyberg/http-client/blob/master/http-client-openssl/Network/HTTP/Client/OpenSSL.hs#L71) closes before SSL_read does not finishes.

This PR fixes the code and adds a test for response-body.
The test's result is below.
```
$ stack test --flag http-client-openssl:test-proxy
http-client-openssl-0.2.1.1: test (suite: spec)


make a TLS connection
make a TLS connection with proxy
compare responses without and with proxy

Finished in 5.4022 seconds
3 examples, 0 failures
````

### DIFF
--- a/http-client-openssl/ChangeLog.md
+++ b/http-client-openssl/ChangeLog.md
@@ -1,3 +1,7 @@
+## 0.2.1.1
+
+* Fix a connection-bug with http-proxy(Previous version closes a connection before reading all respose-data.)
+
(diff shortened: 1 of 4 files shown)


junjihashimoto:
thx!

PR #303, opened 2017-09-25
make the chunked transfer-encoding case-insensitive

cleverca22:


http-client/Network/HTTP/Client/Response.hs
```diff
@@ -87,7 +89,7 @@ getResponse :: ConnRelease
 getResponse connRelease timeout' req@(Request {..}) conn cont = do
     StatusHeaders s version hs <- parseStatusHeaders conn timeout' cont
     let mcl = lookup "content-length" hs >>= readDec . S8.unpack
-        isChunked = ("transfer-encoding", "chunked") `elem` hs
+        isChunked = ("transfer-encoding", mk "chunked") `elem` map (second mk) hs
```
taktoa: Perhaps it would be better to import `Data.CaseInsensitive` qualified, as the [documentation suggests](https://hackage.haskell.org/package/case-insensitive/docs/Data-CaseInsensitive.html)?

PR #308, opened 2017-11-09
Update example at top

FintanH:
The original example wouldn't type check due to the result being `Response L8.ByteString` instead of just `L8.ByteString`. Compiler error here:
```
• Couldn't match type ‘Response L8.ByteString’ with ‘L8.ByteString’
      Expected type: Response L8.ByteString -> IO ()
        Actual type: L8.ByteString -> IO ()
    • In the second argument of ‘(>>=)’, namely ‘L8.putStrLn’
      In the expression: httpLBS "http://example.com" >>= L8.putStrLn
      In an equation for ‘main’:
          main = httpLBS "http://example.com" >>= L8.putStrLn
```

Proposed the change of using `L8.pack . show` to encode the `Response`. Could also do something simpler such `getResponseBody` to just display the body of the response. Not sure which would be better to display for someone arriving to the library.

### DIFF
--- a/http-conduit/Network/HTTP/Simple.hs
+++ b/http-conduit/Network/HTTP/Simple.hs
@@ -14,7 +14,7 @@
 -- > import qualified Data.ByteString.Lazy.Char8 as L8
 -- >
 -- > main :: IO ()
--- > main = httpLBS "http://example.com" >>= L8.putStrLn
+-- > main = httpLBS "http://example.com" >>= L8.putStrLn . L8.pack . show
 --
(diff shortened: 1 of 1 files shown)


snoyberg:
I think `getResponseBody` would be a better choice here, would you mind updating? And thanks for catching this!

FintanH:
Changed it to `getResponseBody` :) No problem, it came up in the fp slack so wanted to make sure it wasn't an issue for any further new comers 😁

PR #319, opened 2018-01-18
Correct lower-bounds

phadej:
- base: `Prelude.catch` and `Control.Exception.catch` are different in
  `base-4.5`
- containers: There are uses of `Data.Map.Strict`, present in
  `containers-0.5.*` only
- stm: `swapTVar` is there only from `stm-2.3`

looks like GHC-7.6.3 is commented out on travis, please bump the lower-bound for base (to `base >= 4.7`) if you don't plan to support GHC-7.6 in the future releases

I made this revision to latest release on Hackage: https://hackage.haskell.org/package/http-client-0.5.8/revisions/


snoyberg:
Thanks!

PR #321, opened 2018-01-31
Make search API of github library work again

tkvogt:
Using recent [patch to http-types](https://github.com/aristidb/http-types/pull/84) to fix [github issue #262](https://github.com/phadej/github/issues/262).

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,3 +1,7 @@
+## 0.5.10
+
+* New function to partial escape query strings
+
 ## 0.5.9
 
 * Add `Semigroup` instances for GHC 8.4 [#320](https://github.com/snoyberg/http-client/pull/320)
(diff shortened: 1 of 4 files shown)


snoyberg:
I don't follow how the added function address the title in this PR. Also, can you include a change of entry and bump the version in the cabal file?

tkvogt:
I forgot to mention this [issue](https://github.com/aristidb/http-types/issues/74), where it is explained.
Github library uses http-client which uses http-types. http-types now allows to explicitly say which part of the query to encode. If you merge this patch, it is possible to use '+' or ':' in a search keyword, or what is more important to prevent '+' and ':' to be URL encoded and destroy a query like this:

`?q=a%2Bcreated%3A2012-12-07..2012-12-17`

snoyberg:
OK, got it, this makes sense now. There's only one problem still: Stackage has been unable to upgrade to the newer version of http-types because so many packages are incompatible with it. I think it would make sense to make this new function's presence conditional on the version of http-types, instead of adding a lower bound.

snoyberg:
This removes the lower bound, but doesn't make inclusion of the function conditional on the http-types version, meaning that this will simply break for older versions of http-types (Travis and AppVeyor will likely demonstrate this).

snoyberg:
Thanks!

PR #324, opened 2018-02-01
Keep original TLS settings when creating a manager using newTlsManagerWith

### LINKED ISSUE #289: function "newTlsManagerWith" does not take into account TLSSettings
I just tried to test functions from my small project in ghci, and have found that function "newTlsManagerWith" from Network.HTTP.Client.TLS module  does not take into account TLSSettings. 

my actions in ghci: 
> tlsset = TLSSettingsSimple True False True  -- i disable certificate checking
> mnsett = mkManagerSettings tlsset Nothing
> manager2  <- newTlsManagerWith mnsett
> response <- Network.HTTP.Client.httpLbs req2 manager2 

and i got TLS error : 

"InternalException (HandshakeFailed (Error_Protocol ("certificate rejected: [NameMismatch \"192.168.5.46\",LeafNotV3]",True,CertificateUnknown)))" 

But when i tried to use function "newManagerSettings" from "Network.HTTP.Client.Conduit" it worked !

> manager2  <- newManagerSettings mnsett
> response <- Network.HTTP.Client.httpLbs req2 manager2 

And i got proper response without any errors. 
Version of http-client-tls is 0.3.5.1 from lts-8.22

I hope my report helps to someone =)

tfausak:
This fixes #289. 

I hoped that `ManagerSettings` would have a field for its `TLSSettings`, but unfortunately it doesn't. Instead of adding such a field and using it here, I chose to copy over the fields that are derived from the `TLSSettings` instead. This way I don't have to change the core `ManagerSettings` data type. The downside is that if any more fields depend on the `TLSSettings`, it's unlikely that `newTlsManagerWith` will also end up being updated. 

:pager: @ivan-m

### DIFF
--- a/http-client-tls/ChangeLog.md
+++ b/http-client-tls/ChangeLog.md
@@ -1,3 +1,6 @@
+* [#289](https://github.com/snoyberg/http-client/issues/289):
+  Keep original `TLSSettings` when creating a `Manager` using `newTlsManagerWith`.
+
 ## 0.3.5.1
 
 * Also catch TLSError exceptions [#273](https://github.com/snoyberg/http-client/pull/273)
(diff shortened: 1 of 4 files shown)


snoyberg:
Thanks!

PR #328, opened 2018-02-17
Add missing `build-tools` dependency specification

hvr:
also provides cabal project configuration file

### DIFF
--- a/http-client/http-client.cabal
+++ b/http-client/http-client.cabal
@@ -77,6 +77,7 @@ test-suite spec
   hs-source-dirs:      test
   default-language:    Haskell2010
   other-modules:       Network.HTTP.ClientSpec
+  build-tools:         hspec-discover
   build-depends:       base
                      , http-client
                      , hspec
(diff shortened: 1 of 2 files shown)


http-client/http-client.cabal
```diff
@@ -77,6 +77,7 @@ test-suite spec
   hs-source-dirs:      test
   default-language:    Haskell2010
   other-modules:       Network.HTTP.ClientSpec
+  build-tools:         hspec-discover
```
snoyberg: What's the motivation for adding this? The builds have been passing with both `cabal-install` and `Stack` without them. I _think_ this is for `new-build`. If so, please add a comment explaining this motivation, otherwise it's likely that future changes may remove it. It would also be a good idea to add a new Travis stanza which includes `new-build` to prevent a change accidentally breaking `new-build` in the future.

snoyberg:
Ping, just confirming you saw my request for changes.

snoyberg:
Closing due to lack of response.

http-client/http-client.cabal
hvr: It's documented [here](http://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-build-tools)

Issue #327, opened 2018-02-17
compile failure


hvr:
see https://ci.appveyor.com/project/hvr/hackage-security/build/12/job/t4l3a6vkcss1yi7h#L1638

snoyberg:
Thanks for the report. I'm unable to reproduce, but adding `FlexibleContexts` is a relatively benign change. I've pushed 24488e4a403b298f34d996fd11ae883bb29ecdb9, can you confirm if this fixes the problem for you?

snoyberg:
From other PRs, it seems like this was in fact resolved by the commit I pushed, closing.

PR #329, opened 2018-02-17
Improved Appveyor CI script

hvr:
This improved CI allows to reproduce and expose the issue from #327, and possibly other issues with a couple of tests in the `spec-nonet` test-suite.

You're welcome


snoyberg:
Ping, just wondering if you had any questions about my requests for changes here.

PR #331, opened 2018-02-18
Addressed warnings found by running `stack test --pedantic`.

naushadh:
Stumbled upon some warnings and thought I'd submit a patch.

I'm not versed enough with travis or cabal to emulate `stack test --pedantic` to make this an automatic CI level check.


naushadh:
CI failures appear to be the same as the ones in HEAD.

snoyberg:
I need to look into those failures. Thanks for the PR.

PR #334, opened 2018-03-01
Switch out `base64-bytestring` for `memory`

Porges:
> It's probably worth switching over anyway.  — @snoyberg

I'm not sure what the policy is here on upper bounds since there are some upper bounds but not many. I went with the predominant style.


Porges:
Note that existing test covers the change in behaviour (confirmed by replacing with `S8.empty`).

Porges:
Also: do you bump versions on the PR branch?

snoyberg:
Looks great! Yes, we typically bump version numbers in PRs, as well as add notes to the ChangeLog. Would you mind updating that here?

Porges:
I think this is done but the Travis build is very flaky

PR #341, opened 2018-03-25
Fix non-TLS connections going though proxy

ocheron:
Looking at source code from [0.5.7.1](https://github.com/snoyberg/http-client/blob/http-client/0.5.7.1/http-client/Network/HTTP/Client/Manager.hs) (notably functions `getConn` and `getConnDest`) this should restore the previous behavior.

The part I left is the modification of pooling key from commit 8c24f7dbf7c083ece6cc467b3a4fcc35e4727a4e. I'm not sure if that one is necessary.

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,6 +1,7 @@
 ## 0.5.12
 
 * Added `requestFromURI` and `requestFromURI_` functions.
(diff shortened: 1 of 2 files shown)


snoyberg:
Looks good! Could you add a ChangeLog entry and minor version bump?

ocheron:
0.5.12 does not seem to be released yet so I added to the existing entry.

PR #342, opened 2018-03-28
Removed unused extensions

Anabra:
Removed most unused extensions from every module. This has been done using the OrganizeExtensions refactoring of the Haskell-Tools framework. See http://www.haskelltools.org/


Anabra:
Some tests didn't succeed in the original code. Since `stack test` didn't run them, I had to execute them manually through `stack ghci$ :main`. 

The failing tests were:
-  http-client/http-conduit/test/main.hs (FAILED)
-  http-client/integration-test.hs (DIDN'T COMPILE)
-  http-client/http-conduit/test.hs (DIDN'T COMPILE)

snoyberg:
Thanks!

PR #344, opened 2018-04-08
Add signature generation with hmac-sha algorithms

rtentser:


http-conduit/http-conduit.cabal
```diff
@@ -33,6 +33,7 @@ library
                  , http-client-tls       >= 0.3     && < 0.4
                  , mtl
                  , unliftio-core
+                 , SHA
```
snoyberg: Since we already depend on cryptonite, which to my knowledge provides this functionality, I'd rather depend on that library here and not increase the dependency footprint. Is there a problem with that approach?

PR #346, opened 2018-04-12
Request bodies in LBS

rtentser:
Sometimes you need to use a request body as a string. But work with RequestBody doesn't simple.
So i add in http-client a function to get lazy bytestring from it.
And i add in http-conduit (Network.HTTP.Simple) a function that takes request and use http-client function to do the same.
They work only with RequestBodyLBS constructor.

### DIFF
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -171,6 +171,7 @@ module Network.HTTP.Client
     , streamFile
     , observedStreamFile
     , StreamFileStatus (..)
+    , requestBodyLBS
       -- * Response
     , Response
     , responseStatus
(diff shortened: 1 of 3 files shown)


snoyberg:
If you need this functionality, it's possible to simply pattern match in your code. I'm not in favor of adding partial functions like this to the library.

rtentser:
I had changed the result type of requestBodyLBD to Maybe ByteString so it is Just LBS if constructor is supported or Nothing otherwise. So now it isn't a partial function.

And i add support for RequestBodyBS constructor.

snoyberg:
I'm still not in favor of this addition: it's unclear why this is better than just letting people pattern match on the value.

rtentser:
> why this is better than just letting people pattern match on the value.

It simpler. And i need it to sign POST requests in https://github.com/snoyberg/http-client/pull/344 right. I don't think you'll be in favor if i'll pattern match there.

Issue #304, opened 2017-09-28
Decent way to use `parseRequest` with checked responses?


mgsloan:
The docs for `parseRequest` indicate that its `checkResponse` won't do anything.  I looked around for a function to add the same `checkResponse` as used by `parseUrlThrow`, but couldn't find one.  I'm thinking it makes sense to:

* Add `setRequestCheckStatus` (opposite of `setRequestIgnoreStatus`), which adds the same status check as `parseUrlThrow`.  Use this function in the definition of `parseUrlThrow`.
* Add `parseRequestThrow`, which also uses `setRequestCheckStatus`.

Thoughts?

mgsloan:
I suppose when using `Network.HTTP.Client`, the `checkResponse` field is available, though the haddocks make it look like a normal function.   However, even in that case, the default status check doesn't seem to be available as a function.

Note that I'm mainly running into this in the context of https://www.stackage.org/haddock/lts-9.6/http-conduit-2.2.3.2/Network-HTTP-Simple.html , which should also export functions that support this.

snoyberg:
I'm +1 on setRequestCheckStatus. I'm very weakly -1 on parseRequestThrow,
just because the story around these functions is already sufficiently long
and annoying.

mgsloan:
Cool, sounds good!  PR incoming

Perhaps `Network.HTTP.Simple` should also re-export `parseUrl` / `parseUrlThrow`?  Currently using that API alone there's no way to use the default status check

snoyberg:
Hmm.... OK, now I'm not sure. If we're talking about just the
Network.HTTP.Simple module, your first recommendation (parseRequestThrow)
sounds better to me for consistency. It's the other modules that have the
backwards compatibility concerns where I'd be worried about adding the new
function name.

PR #348, opened 2018-04-21
Add more functions for throwing StatusCodeException #304

mgsloan:
I re-encountered the issue described in #304 , so figured it was definitely time to address this in the library.

Not sure about only exporting `parseRequestThrow` / `parseRequestThrow_` from the simple API.  Also might make sense to re-export `throwErrorStatusCodes`

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,3 +1,8 @@
+## 0.5.13
+
+* Adds `setRequestCheckStatus` and `throwErrorStatusCodes` functions.
+  See [#304](https://github.com/snoyberg/http-client/issues/304)
+
 ## 0.5.12.1
 
 * Make the chunked transfer-encoding detection case insensitive
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -26,6 +26,7 @@ module Network.HTTP.Client.Request
     , needsGunzip
     , requestBuilder
     , setRequestIgnoreStatus
+    , setRequestCheckStatus
     , setQueryString
 #if MIN_VERSION_http_types(0,12,1)
     , setQueryStringPartialEscape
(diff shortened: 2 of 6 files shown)


http-client/ChangeLog.md
```diff
@@ -1,3 +1,8 @@
+## 0.5.12.2
```
snoyberg: Since this is adding a new export, the version should be 0.5.13.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -77,24 +80,33 @@ parseUrl :: MonadThrow m => String -> m Request
...
+-- @since 0.5.12.2
+throwErrorStatusCodes :: MonadIO m => Response BodyReader -> m ()
+throwErrorStatusCodes res = do
+    let W.Status sci _ = responseStatus res
+    if 200 <= sci && sci < 300
+        then return ()
+        else liftIO $ do
+            chunk <- brReadSome (responseBody res) 1024
+            let res' = fmap (const ()) res
+            throwHttp $ StatusCodeException res' (L.toStrict chunk)
```
snoyberg: We should not export this function, due to the unwrapped call to `throwHttp`. `throwHttp` uses the hidden data type `HttpExceptionContentWrapper`, which is always unwrapped by this library. If we expose this function, there are no guarantees that will happen.

mgsloan: I see, good catch!  I've modified this function to take `Request` as an argument and directly throw a `HttpExceptionRequest`.

http-client/ChangeLog.md
mgsloan: :+1: fixed

mgsloan:
@snoyberg Thanks for reviewing, I think I've addressed your feedback in the new commit.

snoyberg:
Looks like the code doesn't compile due to a syntax error.

mgsloan:
Doh! Fixed

PR #352, opened 2018-06-05
Add a public function to use `Connection` by the user.

igrep:
I'm trying to create a websocket client library supporting websocket
connection via an HTTP proxy server.

To implement the complex features such as referring `http_proxy` environment
variable, I want to combine the http-client package with the websockets package
to reuse http-client's features related to proxy, TLS and so on.

The `withProxiedConnection` function enables that by providing a `Connection`
object configured for `http_proxy` and TLS to the library user while closing
the connection safely.

(Update) Note:

- Here's a sample code to connect by WebSocket with the `withProxiedConnection`: https://github.com/snoyberg/http-client/blob/926542559f46cfffd372d530f464d2dadf2a2033/http-client-tls/sample/sample.hs
- Related reddit thread: https://www.reddit.com/r/haskell/comments/8nh6ud/preventing_an_object_from_getting_gced_and/

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -11,7 +11,7 @@ module Network.HTTP.Client.Core
     , responseClose
     , httpRedirect
     , httpRedirect'
-    , withProxiedConnection
+    , withConnection
     ) where
 
 import Network.HTTP.Types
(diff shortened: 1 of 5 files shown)


http-client/Network/HTTP/Client/Core.hs
```diff
@@ -270,3 +271,11 @@ httpRedirect' count0 http' req0 = go count0 req0 []
 -- Since 0.1.0
 responseClose :: Response a -> IO ()
 responseClose = runResponseClose . responseClose'
+
+-- | Perform an action using a @Connection@ acquired from the given @Manager@,
+--   considering the proxy server.
```
snoyberg: A few points:

1. Is there anything proxy-specific about this code?
2. Can you elaborate in the Haddocks here what the use case is for this, and in particular than normal usage of the library for HTTP requests wouldn't need this function?
3. Please add a minor version bump to the cabal file, add a `@since` comment here, and update the ChangeLog.

igrep: I've dropped words related to proxy from the doc and name. How about this?

igrep:
@snoyberg ping.

snoyberg:
I've been at a conference and traveling for the past two weeks, I'm still catching up.

PR #356, opened 2018-06-19
Add requestMaxAttempts feature

centromere:
This value dictates how many attempts will be made to process a Request.
A retry will only occur if managerRetryableException retruns True.
A value of Nothing represents an infinite number of retries.

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -125,12 +125,7 @@ httpRaw' req0 m = do
     case ex0 of
         Left se | attemptsRemaining req0 && mRetryableException m se ->
             httpRaw' (decrementMaxAttempts req0) m
-        Left se -> do
-            case fromException se of
-                Just (hecw :: HttpExceptionContentWrapper) ->
-                    throwHttp . MaxAttempts . unHttpExceptionContentWrapper $ hecw
-                Nothing ->
-                    throwHttp . MaxAttempts . InternalException $ se
+        Left se -> throwIO se
         Right r -> return r
 
   where
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -136,7 +136,7 @@ httpRaw' req0 m = do
   where
     decrementMaxAttempts :: Request -> Request
     decrementMaxAttempts req =
-        req { requestMaxAttempts = flip (-) 1 <$> requestMaxAttempts req }
+        req { requestMaxAttempts = subtract 1 <$> requestMaxAttempts req }
 
     attemptsRemaining req =
         case requestMaxAttempts req of
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -14,6 +14,7 @@ module Network.HTTP.Client.Core
     , withConnection
     ) where
 
+import Control.Monad (when)
 import Network.HTTP.Types
 import Network.HTTP.Client.Manager
 import Network.HTTP.Client.Types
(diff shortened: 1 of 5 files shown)


http-client/Network/HTTP/Client/Core.hs
```diff
@@ -85,41 +86,63 @@ httpRaw'
...
-        Right res -> case cookieJar req' of
-            Just _ -> do
-                now' <- getCurrentTime
-                let (cookie_jar, _) = updateCookieJar res req now' cookie_jar'
-                return (req, res {responseCookieJar = cookie_jar})
-            Nothing -> return (req, res)
   where
+    decrementMaxAttempts :: Request -> Request
+    decrementMaxAttempts req =
+        req { requestMaxAttempts = flip (-) 1 <$> requestMaxAttempts req }
```
snoyberg: `subtract 1 <$> ...` would probably be more idiomatic

centromere:
CI is failing because `HttpException` has been [augmented](https://github.com/snoyberg/http-client/pull/356/files#diff-496ce9ecd16e88ded74f31e4b2bed49fR237) with a value to indicate that the maximum number of attempts has been reached. I could change the unit tests, but I am not certain that that's the best solution.

Perhaps it would be best to carve out a special case for `requestMaxAttempts` == Just 1 and not wrap any exceptions that occur. What do you think?

snoyberg:
Regaining the original exception values would definitely be important, otherwise this would be a subtly breaking change that could silently break many code bases.

centromere:
I'm not sure why AppVeyor has failed while Travis has succeeded.

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -9,6 +9,7 @@ module Network.HTTP.Client.Types
     , StatusHeaders (..)
     , HttpException (..)
     , HttpExceptionContent (..)
+    , HttpExceptionContentWrapper
```
snoyberg: It looks like this isn't actually being used

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -547,6 +548,18 @@ data Request = Request
...
+    -- ^ The 'Request' will be retried if all the following conditions are met:
+    --
+    -- * this value is greater than 1 or @Nothing@,
+    -- * the exception is retryable as defined by 'managerRetryableException'.
+    --
+    -- Note: Exceptions thrown via 'checkResponse' will not trigger a retry.
+    --
+    -- Default: Just 1
+    --
+    -- @since
```
snoyberg: Can you add a minor version bump, fill in this `@since`, and add a note to the ChangeLog?

http-client/Network/HTTP/Client/Core.hs
```diff
@@ -85,41 +86,58 @@ httpRaw'
...
+        -- exceptions in both.
+        ex <- try $ do
+            cont <- requestBuilder (dropProxyAuthSecure req) (managedResource mconn)
+
+            getResponse timeout' req mconn cont
+
+        case ex of
+            -- Connection was reused, and might have been closed. Try again
+            Left (e :: SomeException) -> do
+                when (managedReused mconn) $
```
snoyberg: IIUC, you're treating the case where we reused a connection as a retry attempt. That would mean that code which previously had retried on a reused connection will no longer do so, because there are no retry attempts left. Is that correct?

http-client/Network/HTTP/Client/Core.hs
```diff
@@ -85,41 +86,58 @@ httpRaw'
...
+            Right res -> case cookieJar req' of
+                Just _ -> do
+                    now' <- getCurrentTime
+                    let (cookie_jar, _) = updateCookieJar res req now' cookie_jar'
+                    return (req, res {responseCookieJar = cookie_jar})
+                Nothing -> return (req, res)
+
+    case ex0 of
+        Left se | attemptsRemaining req0 && mRetryableException m se ->
+            httpRaw' (decrementMaxAttempts req0) m
```
snoyberg: I'm concerned that simply retrying without any kind of backoff logic is the wrong approach. We should instead be doing some kind of exponential delay. I'm also aware that people may want to have customizable logic for this (like the `retry` package), which may make this too complicated to include in http-client itself. However, the current approach of rapid-fire retrying is probably not going to be too useful.

centromere: Ok. I'll close the request.

PR #359, opened 2018-07-08
Support SNI in http-client-openssl

rjmac:
Tell OpenSSL what the host name is before making the connection.

I haven't tried the proxy branch; I don't have an HTTPS proxy to test with.  But it seems likely to be correct, judging from the code.


http-client-openssl/Network/HTTP/Client/OpenSSL.hs
```diff
@@ -38,6 +38,7 @@ opensslManagerSettings mkContext = defaultManagerSettings
                 $ \sock -> do
                     N.connect sock address
                     ssl <- SSL.connection ctx sock
+                    SSL.setTlsextHostName ssl host'
```
snoyberg: Can you add a minor version bump and a changelog entry explaining the change?

rjmac: Certainly!

PR #364, opened 2018-09-15
Omit port for `getUri` when protocol is `https` and `port` is `443`

sol:


sol:
@snoyberg WDYT?

snoyberg:
LGTM. How about also hiding it when insecure and port is 80 for consistency?

sol:
@snoyberg done.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -201,6 +201,10 @@ getUri req = URI
             _ -> S8.unpack $ queryString req
     , uriFragment = ""
     }
+  where
+    port'
+      | secure req && (port req) == 443 || (port req) == 80 = ""
```
snoyberg: Not quite right: this will hide the port for `secure req == True && port req == 80`. Also, personal preference: would you mind either grouping with parents to clarify the operator precedence, or making this two separate guards?

sol: Duh, no sure how I could mess this up.

> or making this two separate guards

That is what I did.

snoyberg:
LGTM! Just waiting on CI

snoyberg:
Thanks!

PR #368, opened 2018-10-24
Wrap HsOpenSSL exceptions into http-client exceptions

alexbiehl:
This wraps exceptions from HsOpenSSL in a similar way to the TLS based backend. Helps streamlining exception handling a lot. Also it fixes an issue with connection reuse in the face of early termination.

This is probably a breaking change as it changes the exceptions being thrown.

### DIFF
--- a/http-client-openssl/ChangeLog.md
+++ b/http-client-openssl/ChangeLog.md
@@ -1,3 +1,8 @@
+## 0.3.0.0
+
+* Wrap HsOpenSSL specific exceptions into http-clients own `HttpExceptionRequest`. This is a breaking change and might need adjustment with respect to exception handling in user code.
+* More robust handling of unexpectedly closed connections
+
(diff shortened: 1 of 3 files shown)


snoyberg:
I don't have any opinion on the changes, as I don't personally use this library. I agree that it should be a major version bump, so including such a bump in the cabal file and an explanation in the ChangeLog would be a good next step.

If you're actively using this library, it might be a good thing to step in as a comaintainer. Would you be interested?

alexbiehl:
Michael, yes we are using this library in production. I am willing to step in as co-maintainer.

alexbiehl:
It's the same as here on github: alexbiehl. Thanks!

snoyberg:
Alright, I've added you on Hackage too. Go ahead a merge/release the changes you want whenever you're ready.

PR #369, opened 2018-10-25
Reestablish completeness check for unlimited body reader

alexbiehl:
This must have gone missing at some point.

### DIFF
--- a/http-client/Network/HTTP/Client/Body.hs
+++ b/http-client/Network/HTTP/Client/Body.hs
@@ -115,9 +115,13 @@ makeUnlimitedReader :: Connection -> IO BodyReader
 makeUnlimitedReader Connection {..} = do
     icomplete <- newIORef False
     return $ do
-        bs <- connectionRead
-        when (S.null bs) $ writeIORef icomplete True
-        return bs
+      complete <- readIORef icomplete
+      if complete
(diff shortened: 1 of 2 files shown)


http-client/Network/HTTP/Client/Body.hs
```diff
@@ -115,9 +115,13 @@ makeUnlimitedReader :: Connection -> IO BodyReader
...
-        bs <- connectionRead
-        when (S.null bs) $ writeIORef icomplete True
-        return bs
+      complete <- readIORef icomplete
+      if complete
+          then return empty
+          else do
+              bs <- connectionRead
+              when (S.null bs) $ writeIORef icomplete True
+              return bs
```
sol, line 124: By now we have:

https://github.com/snoyberg/http-client/blob/ed974a7af8adeb75f4c02b80d3f2bec4857b9f37/http-client/Network/HTTP/Client/Body.hs#L108-L119

Given the `cleanup` action , I think this is even more crucial now.

sol (changes requested):
- [ ] rebase
- [ ] s/completeness/completion/ in commit message ("completeness check" is easily confused with other concepts)
- [ ] test case for that the cleanup action is never called more than once

PR #371, opened 2018-11-15
Reject headers with newlines

ethercrow:
Presumably that would help with my initial problem with req: https://github.com/mrkkrp/req/issues/54

### DIFF
--- a/http-client/Network/HTTP/Client/Headers.hs
+++ b/http-client/Network/HTTP/Client/Headers.hs
@@ -3,6 +3,8 @@
 {-# LANGUAGE ViewPatterns #-}
 module Network.HTTP.Client.Headers
     ( parseStatusHeaders
+    , validateHeaders
+    , HeadersValidationResult (..)
     ) where
 
 import           Control.Applicative            as A ((<$>), (<*>))
@@ -10,6 +12,9 @@ import           Control.Monad
 import qualified Data.ByteString                as S
 import qualified Data.ByteString.Char8          as S8
 import qualified Data.CaseInsensitive           as CI
+import           Data.Char (ord)
+import           Data.Maybe (mapMaybe)
+import           Data.Monoid
 import           Network.HTTP.Client.Connection
 import           Network.HTTP.Client.Types
 import           System.Timeout                 (timeout)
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client/Headers.hs
```diff
@@ -94,3 +99,15 @@ parseStatusHeaders conn timeout' cont
...
+data HeadersValidationResult = GoodHeaders | BadHeaders { _reason :: S.ByteString }
+
+validateHeaders :: RequestHeaders -> HeadersValidationResult
+validateHeaders headers =
+    case mapMaybe validateHeader headers of
+        [] -> GoodHeaders
+        reasons -> BadHeaders (S8.unlines reasons)
+    where
+    validateHeader (k, v)
+        | S8.count '\n' v > 0 = Just ("Header " <> CI.original k <> " has newlines")
```
qrilka: What about using `S8.elem` instead?

ethercrow: Sure, nice catch.

http-client/Network/HTTP/Client/Headers.hs
```diff
@@ -94,3 +99,15 @@ parseStatusHeaders conn timeout' cont
         return (CI.mk $! strip key, strip $! S.drop 1 bs2)
 
     strip = S.dropWhile (== charSpace) . fst . S.spanEnd (== charSpace)
+
+data HeadersValidationResult = GoodHeaders | BadHeaders { _reason :: S.ByteString }
```
snoyberg: Please do not create partial accessors.

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -166,6 +166,7 @@ data HttpExceptionContent
                    -- @since 0.5.0
                    | InvalidHeader S.ByteString
                    -- ^ The given response header line could not be parsed
+                   --   or given request header is not compliant (e.g. has newlines)
```
snoyberg: I don't like abusing an existing constructor for a new meaning.

ethercrow: I understand what you mean, but this particular case seems like a perfect fit, InvalidHeader for invalid headers. Anyway, I don't feel strongly about this.

ethercrow:
I don't understand the test failures.

sol:
> Please also include a minor version bump and a changelog entry.
@snoyberg I think technically this should have been a major version bump, as adding a new constructor is a breaking change, right?

snoyberg:
I think you meant to comment on https://github.com/snoyberg/http-client/pull/377, right?

sol:
No.  I was referring to the added `InvalidRequestHeader` constructor.  Existing code that does pattern matching on `HttpExceptionContent` is now partial.   This is why I assume that this should have been a major version.

Anyway, it's already released.  No action needed.

PR #377, opened 2019-01-07
Allow renderParts to be used in pure code

sol:
I want to use `Network.HTTP.Client.MultipartFormData.renderParts` in pure code.  So what I did is generalize `renderParts` over arbitrary applicative functors.  This allows me to use it with `runIdentity`.

@snoyberg please feel free to bikeshed, anything that enables my use case works for me.

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,11 @@
 # Changelog for http-client
 
+## 0.6.0
+
+* Generalize `renderParts` over arbitrary applicative functors.  One particular
+  use case that is enabled by this change is that now `renderParts` can be used
+  in pure code by using it in combination with `runIdentity`.
+
 ## 0.5.14
 
 * Omit port for `getUri` when protocol is `http` and port is `80`, or when
--- a/http-client/Network/HTTP/Client/MultipartFormData.hs
+++ b/http-client/Network/HTTP/Client/MultipartFormData.hs
@@ -24,6 +24,7 @@ module Network.HTTP.Client.MultipartFormData
     (
     -- * Part type
      Part
+    ,PartM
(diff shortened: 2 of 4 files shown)


snoyberg:
My only real concern is that this is a breaking change... but it seems to be warranted in this case. Can you add a major version bump and a note in the ChangeLog?

http-client/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog for http-client
 
+## 0.6.0
+
+* Generalize `renderParts` over arbitrary applicative functors.  One particular
+  use case that is enabled by this change is than now `renderParts` can be used
```
snoyberg:
```suggestion
  use case that is enabled by this change is that now `renderParts` can be used
```

sol:
@snoyberg thanks for the quick review.

> My  only real concern is that this is a breaking change... but it seems to  be warranted in this case.

This is true.

However, breakage should be minimal.  Specifically I went with

```haskell
type Part = PartM IO

data PartM m = Part ..
```

instead of (what I would usually prefer)

```haskell
type IOPart = Part IO

data Part m = Part ...
```
to minimize breakage.

> Can you add a major version bump and a note  in the ChangeLog?

Done.

snoyberg:
I like the approach, thank you :). I put out a tweet mentioning this:

https://twitter.com/snoyberg/status/1082688452133601282

Unless I get someone screaming at me in the next 12 hours, I'll merge and release this tomorrow.

http-client/ChangeLog.md
sol: fixed.

sol:
@snoyberg thanks a lot!

PR #379, opened 2019-01-13
Add setUriEither to Network.HTTP.Client.Internal

sol:


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -226,9 +227,16 @@ extractBasicAuthInfo uri = do
...
 -- | Validate a 'URI', then add it to the request.
 setUri :: MonadThrow m => Request -> URI -> m Request
-setUri req uri = do
+setUri req uri = either throwInvalidUrlException return (setUriEither req uri)
+  where
+    throwInvalidUrlException = throwM . InvalidUrlException (show uri)
+
+-- | A variant of `setUri` that returns an error message on validation errors,
+-- instead of propagating them with `throwM`.
+setUriEither :: Request -> URI -> Either String Request
```
snoyberg: Can you add a `@since` comment here?

sol: Done.

PR #390, opened 2019-03-08
Add Network.HTTP.Client.Conduit.httpSource

zudov:
As the comment says, this is analogous to `Network.HTTP.Simple.httpSource`, but gets the manager from `MonadReader`.

The main idea is to provide a utility that idiomatically (through `bracketP`) links together the `responseOpen` and `responseClose` in a single conduit.


zudov:
Added a small entry to changelog.

PR #392, opened 2019-03-11
Fix lower bound on http-client in http-conduit

### LINKED ISSUE #391: Incorrect lower bound on http-client in http-conduit
https://hackage.haskell.org/package/http-client-0.6.2/docs/Network-HTTP-Client.html#v:shouldStripHeaderOnRedirect

```
Building library for http-conduit-2.3.6..
[1 of 3] Compiling Network.HTTP.Client.Conduit ( Network/HTTP/Client/Conduit.hs, dist/build/Network/HTTP/Client/Conduit.o )
[2 of 3] Compiling Network.HTTP.Conduit ( Network/HTTP/Conduit.hs, dist/build/Network/HTTP/Conduit.o )

Network/HTTP/Conduit.hs:161:7: error:
    Not in scope: ‘shouldStripHeaderOnRedirect’
    |
161 |     , shouldStripHeaderOnRedirect
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[3 of 3] Compiling Network.HTTP.Simple ( Network/HTTP/Simple.hs, dist/build/Network/HTTP/Simple.o )

Network/HTTP/Simple.hs:98:1: warning: [-Wunused-imports]
    The import of ‘Data.Monoid’ is redundant
      except perhaps to import instances from ‘Data.Monoid’
    To import instances alone, use: import Data.Monoid()
   |
98 | import Data.Monoid
   | ^^^^^^^^^^^^^^^^^^
```

Reason of this build plan is

```
[__3] rejecting: amazonka-core-1.6.1, amazonka-core-1.6.0 (conflict:
http-client==0.6.2, amazonka-core => http-client>=0.4 && <0.6)
```

hasufell:
Fixes #391

This should be accompanied by uploading a hackage revision to unbreak build plans.

### DIFF
--- a/http-conduit/http-conduit.cabal
+++ b/http-conduit/http-conduit.cabal
@@ -29,7 +29,7 @@ library
                  , conduit               >= 1.2
                  , conduit-extra         >= 1.1
                  , http-types            >= 0.7
-                 , http-client           >= 0.5.13  && < 0.7
+                 , http-client           >= 0.6.2   && < 0.7
                  , http-client-tls       >= 0.3     && < 0.4
                  , mtl
                  , unliftio-core


snoyberg:
I don't mind merging this, but a release to Hackage isn't going to fix the situation, at least not reliably. Better bets:

* Don't rely on dependency solving
* Add more constraints to your project to guide it better
* Change this PR to be compatible with old and new http-client using MIN_VERSION CPP

mrkkrp:
I think @hasufell talks about publishing a new revision, not release. Once metadata is updated to reflect the reality I don't see why it wouldn't work. At least, publishing a revision is not hard even if you don't think it's the best way, others may still benefit from it, so why not?

hasufell:
> I don't mind merging this, but a release to Hackage isn't going to fix the situation, at least not reliably. Better bets:

I'm talking about a hackage revision: https://github.com/haskell-infra/hackage-trustees/blob/master/revisions-information.md

Hackage revisions are exactly what this is for and they fix the situation.

> * Don't rely on dependency solving

Dependency solving works perfectly well and we rely on it since quite some time. It only doesn't work well when dependencies in cabal files are wrong.

> * Add more constraints to your project to guide it better

Our project doesn't directly depend on http-conduit, there is no point in doing this.

> * Change this PR to be compatible with old and new http-client using MIN_VERSION CPP

In what way?

snoyberg:
Thanks for clarifying @mrkkrp. I don't do these kinds of revisions, mostly because it's signing up for an unbounded amount of busywork that I have no way of easily testing for actually solving the problem. If you'd like to do the revisions, I can give you Hackage maintainership.

@hasufell Your comments are inconsistent, I'm not going to get into a debate here. I'm not sure what is unclear about my MIN_VERSION comment.

hasufell:
> Your comments are inconsistent

They are not and I also don't want to get into a debate about how to handle our project. We made decisions and these work well. This PR is simply about fixing the .cabal file and uploading a hackage revision.

> I'm not sure what is unclear about my MIN_VERSION comment.

You didn't really explain how to make it compatible with multiple versions. I am not familiar with the codebase or the functions that are used to know how to make it compatible with multiple versions.

snoyberg:
As Mark noticed, "uploading a revision" confused me. I did not understand you meant a Hackage revision, I thought you made a new upload. I was trying to point out other approaches that would actually solve the problem.

The CPP approach is fairly straightforward, and doesn't require any deep knowledge of the codebase. It's about conditionally reexporting the identifier. If that's not clear enough, I don't have time to go into more detail.

hasufell:
> The CPP approach is fairly straightforward, and doesn't require any deep knowledge of the codebase. It's about conditionally reexporting the identifier. If that's not clear enough, I don't have time to go into more detail.

Ok, I'll try that shortly!

PR #396, opened 2019-03-31
Consume trailing whitespace on JSON bodies (fixes #395)

### LINKED ISSUE #395: httpJSON does not reuse connections
I have the following program that fetches three JSON resources from an `https` server:
```
{-# LANGUAGE OverloadedStrings   #-}
{-# LANGUAGE ScopedTypeVariables #-}

import qualified Data.Aeson           as Aeson
import           Network.HTTP.Client
import           Network.HTTP.Simple

main :: IO ()
main = do
  res <- mapM httpLBS urls
  mapM_ (print . getResponseStatusCode) res

urls :: [Request]
urls = [ "https://hacker-news.firebaseio.com/v0/item/19493624.json"
       , "https://hacker-news.firebaseio.com/v0/item/19506044.json"
       , "https://hacker-news.firebaseio.com/v0/item/19493277.json"
       ]
```

I run it like this:
```
$ stack runghc conn.hs 
200
200
200
```

If at the same time I run `tcpdump`, I see only one connection to `hacker-news.firebaseio.com:443`:
```
$ sudo tcpdump 'tcp[tcpflags] & tcp-syn != 0 and host hacker-news.firebaseio.com' 
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on wlp4s0, link-type EN10MB (Ethernet), capture size 262144 bytes
21:11:33.169527 IP pergamon.40692 > 85.97.201.35.bc.googleusercontent.com.https: Flags [S], seq 1754220615, win 29200, options [mss 1460,sackOK,TS val 3198986189 ecr 0,nop,wscale 7], length 0
21:11:33.187130 IP 85.97.201.35.bc.googleusercontent.com.https > pergamon.40692: Flags [S.], seq 4261482672, ack 1754220616, win 60192, options [mss 1380,sackOK,TS val 2233905607 ecr 3198986189,nop,wscale 8], length 0
```

**But** if I replace this line
```
res <- mapM httpLBS urls
```
with
```
(res :: [Response Aeson.Value]) <- mapM httpJSON urls
```
then I get three separate connections, one for each request.

My versions:
```
$ stack ls dependencies | grep http-
http-client 0.5.14
http-client-tls 0.3.5.3
http-conduit 2.3.6.1
http-types 0.12.3
```

snoyberg:
CC @vfaronov, fixes #395

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,12 @@
 # Changelog for http-client
 
+## 0.6.3
+
+* Detect response body termination before reading an extra null chunk
+  when possible. This allows connections to be reused in some corner
+  cases. See
+  [#395](https://github.com/snoyberg/http-client/issues/395)
+
 ## 0.6.2
 
 * Add `shouldStripHeaderOnRedirect` option to `Request` [#300](https://github.com/snoyberg/http-client/issues/300)
--- a/http-client/Network/HTTP/Client/Body.hs
+++ b/http-client/Network/HTTP/Client/Body.hs
@@ -8,7 +8,6 @@ module Network.HTTP.Client.Body
     , brConsume
     , brEmpty
     , constBodyReader
-    , brAddCleanup
     , brReadSome
     , brRead
     ) where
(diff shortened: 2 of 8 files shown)


snoyberg:
I'm not convinced this is a good fix. It would simply ignore plenty of non-whitespace characters after parsing a valid JSON response. As it happens, that's exactly what this function does today, but perhaps we need to be more thorough with checking. It would be interesting to see what other libraries in other languages do for this.

vfaronov:
> It would be interesting to see what other libraries in other languages do for this.

At least in Go and Python, it is typical to read the entire HTTP response into memory first, and then parse it (which succeeds because trailing whitespace is valid in JSON). If you want fusing, you have to arrange for it manually.

Issue #398, opened 2019-04-10
fails with InvalidHeader on url that is supported by curl and wget


joeyh:
http-client fails to download the url https://datahub-khvul4ng.udes.genap.ca/ALL.chr10.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz , with an InvalidHeader exception. The same url can be downloaded by wget, curl, and firefox.

Examining the http response (with curl -v), it is indeed a bit broken:

         < Server: openresty/1.7.10.1
        < Date: Wed, 10 Apr 2019 13:57:10 GMT
        < Content-Type: application/octet-stream
        < Content-Length: 773788987
        < Connection: keep-alive
        < Last-Modified: Mon, 11 Mar 2019 16:35:02 GMT
        < ETag: "5c868e36-2e1f153b"
        < Accept-Ranges: bytes
        < Strict-Transport-Security: max-age=15768000; includeSubdomains;
        < preload
        < X-Frame-Options: DENY
        < X-Content-Type-Options: nosniff

In particular, the "preload" has somehow ended up on its own line, not part of the Strict-Transport-Security header.

Looks like the code that throws the exception is here:

    parseHeader :: S.ByteString -> IO Header
    parseHeader bs = do
        let (key, bs2) = S.break (== charColon) bs
        when (S.null bs2) $ throwHttp $ InvalidHeader bs
        return (CI.mk $! strip key, strip $! S.drop 1 bs2)

Since there's no colon, the snd is null.

In my application I'd rather that http-client applied Postel's law the same way that curl, wget, and firefox do, by not throwing an exception here, but ignoring (or somehow logging a problem with) the malformed header. 

One way to do it would be this, so when there's a missing colon, it still parses a Header, just one with an empty value.

    parseHeader :: S.ByteString -> IO Header
    parseHeader bs = do
        let (key, bs2) = S.break (== charColon) bs
        if S.null bs2
                then return (CI.mk $! strip key, bs2)
                else return (CI.mk $! strip key, strip $! S.drop 1 bs2)

(Of course that could be refactored.)

snoyberg:
If the other HTTP libraries are accepting this, we may as well too. You up for sending a PR after confirming the change addresses your use case?

joeyh:
I've confirmed the change works. Pr coon.

PR #399, opened 2019-04-11
don't fail on a single malformed header from the server

joeyh:
A line in the HTTP response that lacks a ':' caused an exception,
but other commonly used http implementations (curl, wget, firefox)
are robust in what they accept here. So a web server that is
misconfigured with eg, a '\n' in the middle of a header value,
would seem to work ok but break when http-client is used.
Fixed by ignoring such bogus and unparseable lines.

The InvalidHeader constructor is no longer used in http-client, so could
be later removed; it has been left in to avoid this bug fix getting
entangled with an API bump.

Fixes https://github.com/snoyberg/http-client/issues/398

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,11 @@
 # Changelog for http-client
 
+## 0.6.4
+
+* Avoid throwing an exception when a malformed HTTP header is received,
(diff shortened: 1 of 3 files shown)


joeyh:
made those changes

PR #408, opened 2019-07-09
Responses with a status code that signals "no content" are not handled properly through an SSL proxy

albertov:
To reproduce install a https proxy (tried with Hoverfly), set the `https_proxy` environment variable and  run the following script:
```haskell
#!/usr/bin/env stack
-- stack -v runghc --package connection --package http-client --package http-client-tls 

import Prelude

import Network.Connection
import Control.Monad (forM)
import Network.HTTP.Client
import Network.HTTP.Client.TLS


main :: IO ()
main = do

    let (Just req) = parseUrlThrow "https://assertible.com/blog"
    let managerSettingsTls =
          mkManagerSettings
            (TLSSettingsSimple True False False) -- Don't check certificate because proxy may mitm it
            Nothing
    man <- newManager $
      managerSetProxy (proxyEnvironment Nothing) $
      managerSettingsTls

    results <- forM [1..250] $ \_ -> do
        res <- httpLbs req{method="HEAD"} man
        pure (responseStatus res)

    print (length results)
```
The above script, when compiled with the `-threaded` RTS, will randomly fail with:
```
invalid-status-line: HttpExceptionRequest Request {
  host                 = "assertible.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/blog"
  queryString          = ""
  method               = "HEAD"
  rawBody              = False
  redirectCount        = 10
}
 (InvalidStatusLine "0")
```

This PR avoids the problem although it probably isn't the optimal solution as it forces keep-alive connections to be closed. Maybe consuming the (empty) response is a better fix.

### DIFF
--- a/http-client/Network/HTTP/Client/Response.hs
+++ b/http-client/Network/HTTP/Client/Response.hs
@@ -11,6 +11,7 @@ import qualified Data.ByteString.Char8 as S8
 import qualified Data.ByteString.Lazy as L
 import qualified Data.CaseInsensitive as CI
 import Control.Arrow (second)
+import Control.Monad (when, unless)
 
 import Data.Monoid (mempty)
 
@@ -102,6 +103,13 @@ getResponse timeout' req@(Request {..}) mconn cont = do
         -- RFC 2616 section 4.4_1 defines responses that must not include a body
         if hasNoBody method (W.statusCode s) || (mcl == Just 0 && not isChunked)
             then do
+                when isChunked $ do
+                    -- if chunked encoding we need to consume the empty chunk's
+                    -- header or it'll linger if the connection is kept alive on
+                    -- the connection buffer and mess up the next request
+                    -- See https://github.com/snoyberg/http-client/pull/408
+                    bs <- connectionRead conn
+                    unless ( bs == "0\r\n") $ throwHttp InvalidChunkHeaders
                 cleanup True
                 return brEmpty
             else do
--- a/http-client/Network/HTTP/Client/Response.hs
+++ b/http-client/Network/HTTP/Client/Response.hs
@@ -102,7 +102,7 @@ getResponse timeout' req@(Request {..}) mconn cont = do
         -- RFC 2616 section 4.4_1 defines responses that must not include a body
         if hasNoBody method (W.statusCode s) || (mcl == Just 0 && not isChunked)
             then do
-                cleanup False -- The body hasn't actually been consumed
+                cleanup True
                 return brEmpty
             else do
                 body1 <-
--- a/http-client/Network/HTTP/Client/Response.hs
+++ b/http-client/Network/HTTP/Client/Response.hs
@@ -102,7 +102,7 @@ getResponse timeout' req@(Request {..}) mconn cont = do
         -- RFC 2616 section 4.4_1 defines responses that must not include a body
         if hasNoBody method (W.statusCode s) || (mcl == Just 0 && not isChunked)
             then do
-                cleanup True
+                cleanup False -- The body hasn't actually been consumed
                 return brEmpty
             else do
                 body1 <-
--- a/http-client/Network/HTTP/Client/Response.hs
+++ b/http-client/Network/HTTP/Client/Response.hs
@@ -101,15 +101,12 @@ getResponse timeout' req@(Request {..}) mconn cont = do
 
     body <-
         -- RFC 2616 section 4.4_1 defines responses that must not include a body
-        if hasNoBody method (W.statusCode s) || (mcl == Just 0 && not isChunked)
+        -- If the transfer-encoding is chunked then we need to consume the empty
+        -- chunk header or it will linger in the connection buffer and mess up
+        -- the next request in proxied keep-alive connections.
+        -- See https://github.com/snoyberg/http-client/pull/408
+        if not isChunked && (hasNoBody method (W.statusCode s) || mcl == Just 0)
             then do
-                when isChunked $ do
-                    -- if chunked encoding we need to consume the empty chunk's
-                    -- header or it'll linger if the connection is kept alive on
-                    -- the connection buffer and mess up the next request
-                    -- See https://github.com/snoyberg/http-client/pull/408
-                    bs <- connectionRead conn
-                    unless ( bs == "0\r\n") $ throwHttp InvalidChunkHeaders
                 cleanup True
                 return brEmpty
             else do


albertov:
BTW, The repro has been adapted from the one in https://github.com/yesodweb/wai/issues/634 which seems to be the same underlying issue

snoyberg:
Can you provide a bit more information explaining why this is failing specifically for an SSL proxy and how this fixes it?

albertov:
> Can you provide a bit more information explaining why this is failing specifically for an SSL proxy and how this fixes it?

My understanding is that trailing bytes are left in the connection buffer when [hasNoBody](https://github.com/snoyberg/http-client/blob/4bab8a4b001be1ef8cf5f663f23fc079861d457e/http-client/Network/HTTP/Client/Response.hs#L129) returns `True` when the response body is either consumed or not here: https://github.com/snoyberg/http-client/blob/4bab8a4b001be1ef8cf5f663f23fc079861d457e/http-client/Network/HTTP/Client/Response.hs#L103

Since connections to the proxy are left open after issuing `CONNECT`, the next request that reuses the connection to reads these leftover bytes which makes parsing the response's headers fail in https://github.com/snoyberg/http-client/blob/4bab8a4b001be1ef8cf5f663f23fc079861d457e/http-client/Network/HTTP/Client/Headers.hs#L71.

The fix is certainly a kludge since it works around the problem by calling `cleanup False` instead of `cleanup True` to signal that the body hasn't been consumed so that the connection isn't reused. I guess the proper solution would be to consume the response regardless if it's expected to be empty or not to make sure there are no left-over input to consume. Anyway, it still serves as evidence in favor of the hypothesis.

I don't know why it only manifests itself when being an SSL proxy. I guess it's related to the internal buffering that `Network.TLS.recvData` does

snoyberg:
I'm not comfortable merging a kludge like this as-is without further understanding of what is happening here.

albertov:
Fair enough, please consider this as a bug report with a repro. The kludge is presented as a PR because it fixes the issue for us so it might give a clue about what the root cause might be to someone more versed in the implementation. This should have been filed as an issue really.

albertov:
BTW, I failed to mention that the repro executable pasted in the description needs to be linked against the threaded RTS for the bug to manifest itself. This sample `.cabal` file can be used to build it:
```
cabal-version:       >=1.10
name:                invalid-status-line
version:             0.1.0.0
license:             BSD3
build-type:          Simple

executable invalid-status-line
  main-is:             invalid-status-line.hs
  build-depends:       base, connection, http-client, http-client-tls
  default-language:    Haskell2010
  ghc-options: -threaded
```

albertov:
bb0f4bd implements a less-kludge fix that allows the connection to be reused and explicitly consumes the `0\r\n' (ie: empty chunk header) that is left in the buffer as evidence of what's going on. ed88d55bf Implements a non-kludge fix and hopefully explains better in a comment what is happening

snoyberg:
Please see this discussion on Stack Overflow: https://stackoverflow.com/questions/912863/is-an-http-application-that-sends-a-content-length-or-transfer-encoding-with-a-2.

A response that doesn't have a response body (e.g. HEAD response or status code 204) should never be sending a chunked body terminator. I think something else is going on here.

albertov:
I’m aware of that the server shouldn’t be se sending an empty chunk header when the request has no body but in this case the proxy  _is_ sending it (see the repro for evidence) and causing an exception in http-client users. Should the client work around it? I think so, due to https://en.m.wikipedia.org/wiki/Robustness_principle

snoyberg:
No, definitely not. Reading extra data from a connection like that can corrupt an upcoming response shared on the same connection. I still believe there's something more complex at play here.

albertov:
Any idea what else could be going on?

snoyberg:
Not really. Do you have a capture of the decrypted communication between the client and the proxy server?

albertov:
I do have a decrypted dump. It seems that the proxy server is indeed sending a chunk header for an empty block (ie : `0\r\n`, the `30 0d 0a` almost at the end of the response) in the response body with a 204 status code:
image

Here's the `pcap` dump and session keys needed to decrypt it with wireshark (https://packetpushers.net/using-wireshark-to-decode-ssltls-packets/)

[dump.zip](https://github.com/snoyberg/http-client/files/3381773/dump.zip)

albertov:
> Reading extra data from a connection like that can corrupt an upcoming response shared on the same connection.

I'm not sure this is the case. My understanding is that the server wouldn't send more data until another request is made on the same connection, right? The next request on the same connection would happen after reading the extra data so I don't see how it could be reading the data of a response to a request that hasn't been made yet. I may be missing something though.

albertov:
I'm attaching another screenshot of the response from the proxy where the response body is highlighted to make it easier to see.
image

It certainly looks like an empty chunk with it's header `0\r\n\r\n` so it appears the proxy server is not following the spec (filed a bug with the proxy: https://github.com/SpectoLabs/hoverfly/issues/865) so this isn't technically a `http-client` bug. However, maybe `http-client` should be more tolerant to non-conforming proxies/servers, if possible, for robustness?

albertov:
In any case, it seems that we should handle the case of a `HEAD` or a `304` response with `transfer-encoding: chunked`:
>    Transfer-Encoding MAY be sent in a response to a HEAD request or in a
   304 (Not Modified) response (Section 4.1 of [RFC7232]) to a GET
   request, neither of which includes a message body, to indicate that
   the origin server would have applied a transfer coding to the message
   body if the request had been an unconditional GET.  This indication
   is not required, however, because any recipient on the response chain
   (including the origin server) can remove transfer codings when they
   are not needed.

I don't think this handled properly now since `hasNoBody` makes no distinction between `204` responses and `HEAD` responses so a chunked empty body may come along and cause this same `InvalidStatusLine "0"` issue.

Yuras:
Just another (hopefully helpful) data point. Curl was reading chunked body in case of 204 status, but recently they changed code to ignore Transfer-Encoding in such cases. At least that is how I understand the code [here](https://github.com/curl/curl/blob/ac3d19a26de79d9ec62d20a1a0b645d7ddcc40fd/lib/http.c#L3787) and [here](https://github.com/curl/curl/blob/ac3d19a26de79d9ec62d20a1a0b645d7ddcc40fd/lib/http.c#L3903). See also curl/curl#3977 and curl/curl#3968

snoyberg:
I'm not seeing a strong argument about changing behavior here in http-client to be honest. Adding additional calls to `read` can result in unnecessary blocking, and writing a client in such a way to expect misbehaving servers isn't something I'm happy about. It's one thing to write a client to accept unambiguous but incorrect input. This is ambiguous and incorrect.

albertov:
What about HEAD or 304 responses with a chunked empty body? These are allowed per spec and will suffer the same fate as 204s as it is now. With this patch they’ll be handled properly and we get misbehaving 204s for “free” (keep in mind the read-body path is only taken if the response is chunked)

snoyberg:
I think we're understanding things very differently. My understanding is that, if a server says that the transfer-encoding is chunked for any of these responses, they are required by the spec to _not_ send a `0\r\n` afterwards. Reading for that content eagerly seems like a violation of the spec on the part of http-client.

albertov:
> I think we're understanding things very differently. My understanding is that, if a server says that the transfer-encoding is chunked for any of these responses, they are required by the spec to _not_ send a `0\r\n` afterwards. Reading for that content eagerly seems like a violation of the spec on the part of http-client.

Agree. FYI the thirdparty bug that motivated this issue has been fixed https://github.com/SpectoLabs/hoverfly/issues/865. Closing this, thanks!

PR #429, opened 2020-03-04
Add version constraints to resourcet and unliftio-core

adithyaov:
* This is to make it work with ghc < 0.8 when using cabal
* Adding a constraint on resourcet isn't required but added for
  additional clarity


snoyberg:
I don't generally accept these kinds of PRs since it's too much complexity that shouldn't exist. You should be locking version numbers, especially for older GHCs.

adithyaov:
> You should be locking version numbers, especially for older GHCs.

I see. Ideally, we would like to stop supporting lower versions of ghc. Versioning in cabal is indeed complex.

PR #430, opened 2020-03-09
Fix Win32-2.8 compat.

angerman:
It appears as if the `regQueryValue` was fixed in the Win32-2.8 series: https://github.com/haskell/win32/commit/1c46705d72341ac608c58f29b61ceadcca7c7440


snoyberg:
Thanks!

PR #431, opened 2020-03-26
expose Network.HTTP.Client.Types

BebeSparkelSparkel:


snoyberg:
There's no motivation provided for doing this, and there's plenty of motivation not to do this.

BebeSparkelSparkel:
It would be very nice to have access to the Request and Response constructors to make convenience functions. Would you be open to exposing those?

snoyberg:
These are already exposed in the `.Internal` module: https://www.stackage.org/haddock/nightly-2020-03-26/http-client-0.6.4.1/Network-HTTP-Client-Internal.html#t:Request

BebeSparkelSparkel:
Thanks

PR #432, opened 2020-05-07
http-client-openssl: Set SNI hostname to serverName when using https_proxy

arianvp:
Setting it to `host'` here is wrong. that's the address of the `https_proxy`, not the server we're trying to make a request to.

Instead, it should be set to `serverName`.

Now the behaviour is the same as in `http-client-tls` which does not have this bug

Fixes https://github.com/wireapp/echo-bot/issues/45#issuecomment-625359416


arianvp:
will do monday

PR #438, opened 2020-07-19
http-conduit: Fix tests on big-endian systems

iliastsi:
Using the hard-coded `0x0100007f` value as `hostAddress` to represent
`127.0.0.1` does not work on big-endian systems. Use the
`tupleToHostAddress` function instead, which correctly computes the
hostAddress value.


snoyberg:
CI failures are unrelated. Thanks!

PR #440, opened 2020-08-16
Add support for TLS offload via proxy

A1kmm:
For server applications that make outbound HTTP requests as a client, it is often preferrable to use a local forward proxy for caching and TLS offload.

For this use case, it makes more sense to send the request directly to the proxy in plain text, with an `https://` URL, so that the proxy can cache the response without needing to be configured to generate self-signed certificates to intercept requests through CONNECT.

This PR adds a new non-default setting to use this mode of connecting to proxies.

### DIFF
--- a/http-client/Network/HTTP/Client/Headers.hs
+++ b/http-client/Network/HTTP/Client/Headers.hs
@@ -12,9 +12,7 @@ import           Control.Monad
 import qualified Data.ByteString                as S
 import qualified Data.ByteString.Char8          as S8
 import qualified Data.CaseInsensitive           as CI
-import           Data.Char (ord)
 import           Data.Maybe (mapMaybe)
-import           Data.Monoid
 import           Network.HTTP.Client.Connection
 import           Network.HTTP.Client.Types
 import           System.Timeout                 (timeout)
--- a/http-client/Network/HTTP/Client/Headers.hs
+++ b/http-client/Network/HTTP/Client/Headers.hs
@@ -13,6 +13,7 @@ import qualified Data.ByteString                as S
 import qualified Data.ByteString.Char8          as S8
 import qualified Data.CaseInsensitive           as CI
 import           Data.Maybe (mapMaybe)
+import           Data.Monoid
 import           Network.HTTP.Client.Connection
 import           Network.HTTP.Client.Types
 import           System.Timeout                 (timeout)
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -17,6 +17,7 @@ module Network.HTTP.Client.Manager
     , proxyEnvironmentNamed
     , defaultProxy
     , dropProxyAuthSecure
+    , useProxySendRequestMode
     ) where
 
 import qualified Data.ByteString.Char8 as S8
@@ -216,7 +217,7 @@ connKey req =
             | secure req -> simple CKSecure
             | otherwise -> simple CKRaw
         Just p
-            | secure req -> CKProxy
+            | secure req && proxySecureMode req == ProxySecureConnect -> CKProxy
                 (proxyHost p)
                 (proxyPort p)
                 (lookup "Proxy-Authorization" (requestHeaders req))
@@ -285,6 +286,16 @@ noProxy = ProxyOverride $ const $ return $ \req -> req { proxy = Nothing }
 -- Since 0.4.7
 useProxy :: Proxy -> ProxyOverride
 useProxy p = ProxyOverride $ const $ return $ \req -> req { proxy = Just p }
+
+-- | Send secure requests to the proxy in plain text rather than using CONNECT,
+-- regardless of the value in the @Request@.
+--
+-- Since 0.7.2
+useProxySendRequestMode :: ProxyOverride -> ProxyOverride
+useProxySendRequestMode (ProxyOverride existingOverride) = ProxyOverride $
+  \secure -> do
+    f <- existingOverride secure
+    return $ f . (\req -> req { proxySecureMode = ProxySecureSendRequest })
 
 -- | Get the proxy settings from the default environment variable (@http_proxy@
 -- for insecure, @https_proxy@ for secure). If no variable is set, then fall
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -17,7 +17,7 @@ module Network.HTTP.Client.Manager
     , proxyEnvironmentNamed
     , defaultProxy
     , dropProxyAuthSecure
-    , useProxySendRequestMode
+    , useProxySecureWithoutConnect
     ) where
 
 import qualified Data.ByteString.Char8 as S8
@@ -211,21 +211,23 @@ getConn req m
     connkey = connKey req
 
 connKey :: Request -> ConnKey
-connKey req =
-    case proxy req of
-        Nothing
-            | secure req -> simple CKSecure
-            | otherwise -> simple CKRaw
-        Just p
-            | secure req && proxySecureMode req == ProxySecureConnect -> CKProxy
-                (proxyHost p)
-                (proxyPort p)
-                (lookup "Proxy-Authorization" (requestHeaders req))
-                (host req)
-                (port req)
-            | otherwise -> CKRaw Nothing (proxyHost p) (proxyPort p)
-  where
-    simple con = con (hostAddress req) (host req) (port req)
+connKey req@Request { proxy = Nothing, secure = False } =
+  CKRaw (hostAddress req) (host req) (port req)
+connKey req@Request { proxy = Nothing, secure = True  } =
+  CKSecure (hostAddress req) (host req) (port req)
+connKey Request { proxy = Just p, secure = False } =
+  CKRaw Nothing (proxyHost p) (proxyPort p)
+connKey req@Request { proxy = Just p, secure = True,
+                      proxySecureMode = ProxySecureWithConnect  } =
+  CKProxy
+    (proxyHost p)
+    (proxyPort p)
+    (lookup "Proxy-Authorization" (requestHeaders req))
+    (host req)
+    (port req)
+connKey Request { proxy = Just p, secure = True,
+                  proxySecureMode = ProxySecureWithoutConnect  } =
+  CKRaw Nothing (proxyHost p) (proxyPort p)
 
 mkCreateConnection :: ManagerSettings -> IO (ConnKey -> IO Connection)
 mkCreateConnection ms = do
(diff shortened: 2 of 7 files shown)


snoyberg:
For comparison purposes, do you have an example of an HTTP client library in another language that offers a similar feature?

A1kmm:
The lower level Python3 built-in http.client library supports this. For example, in Python3:
```python
from http.client import HTTPConnection
conn = HTTPConnection('localhost', 3128)
conn.connect()
conn.request('GET', 'https://google.com/')
```

results in the following being sent to localhost:3128:

```
GET https://google.com/ HTTP/1.1
Host: google.com
Accept-Encoding: identity

```

Many other lower level libraries in other languages (e.g. `http4s`) don't directly implement automatic proxy connection, but leave it up to people building the next layer to decide which host to connect to and what URL (absolute or relative) to request, which allows the user of the library to decide whether to use `CONNECT` or send the request in plain text to the proxy. Higher level libraries are generally more opinionated about proxies.

I would argue that http-client aims to be both a lower level and a higher level HTTP client library, so it should provide enough flexibility to support this use case in the lower level APIs, but not necessarily expose it to the more opinionated higher level APIs.

snoyberg:
I'm having a hard time following this. The original description you provide sounds like a standard HTTP proxy setup, and that's already supported. The terms you're using all add up for that. Furthermore, the example you give from Python sounds like that too. I'm not understanding what use case you're going for that's different than a normal HTTP proxy case.

A1kmm:
There are two ways to use a proxy server to connect to an HTTP server using TLS (i.e. to make an HTTPS connection). Both of them a RFC compliant, but they have different pros and cons:

* HTTP CONNECT: Client connects to the proxy server, and uses the CONNECT method on the proxy to get the proxy to open a connection to the server. The client then starts speaking TLS over the connection via the proxy to provide a secure connection to the server. The pros of this method are that the client, rather than the proxy, has control over the TLS connection (so, for example, the client gets to choose which CAs to trust), and the proxy itself cannot see the contents of the HTTP requests and responses. The cons are that the proxy server cannot (without a workaround) see the contents of the request and make policy decisions, and it cannot cache the response. This mode is generally the right approach for browsers and other client software connecting to a potentially untrusted proxy on behalf of an end user.
* Sending the request to the proxy in plain text (or over a TLS connection between the client and the proxy), with the full absolute URL including scheme and host in the request. The proxy then makes a TLS connection to the server and sends the request to the server. This is identical to how a proxy for HTTP works, except for a TLS protected connection from the proxy to the server. The pros of this approach are that the proxy server can cache the response (if the headers allow it to), and can make policy decisions based on the request. The cons are that the proxy can see the full text of the request and response. This mode is most useful if the client and proxy are part of a solution deployed to a production environment (with the proxy and client on the same host or network), where the client fully trusts the proxy, and the purpose of the proxy is to separate concerns, offload security policy decisions, and enhance production performance.

Some proxies do support a feature where they will intercept CONNECT requests by performing a man-in-the-middle attack to impersonate the server to the client (using a trusted CA certificate that has to be installed into the clients), allowing the proxy to cache requests. However, for the use case of a proxy and client on the same host or network, that is wasteful (a completely architecturally unnecessary TLS session setup) and unnecessarily complex to set up. So supporting the second mode, where the client doesn't need to implement TLS at all, is more efficient for that type of application,

http-client does not currently support the second mode of connection:
* If you try to use it to manually connect to the proxy without using http-client's built in proxy support (using the same approach you would use in Python3 or http4s etc...), it won't send the schema and hostname: https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Request.hs#L471 (`proxy req == Nothing`), and therefore won't be accepted by a normal proxy server.
* If you try to use the built in proxy support to connect to an https URL, then http-client tries to use the CONNECT method: https://github.com/snoyberg/http-client/blob/c1c6ffa9e7bf3556f8577b12fc2ccb3e92b080a4/http-client/Network/HTTP/Client/Manager.hs#L240

This PR keeps the current CONNECT based approach as the default, but gives people access to the alternative approach for use cases where it is more appropriate.

snoyberg:
Can you point out where in the RFC the second proxy method is described?

A1kmm:
[RFC 7230 section 5.3.2](https://tools.ietf.org/html/rfc7230#section-5.3.2) is generic enough to cover both standard HTTP proxy requests and the second method of making an HTTPS request. Per the RFC, if the client is not making a CONNECT request (i.e. not using the first method), then the client must send the absolute-URI. Note that absolute-URI is defined by reference to https://tools.ietf.org/html/rfc3986#section-4.3. RFC7230 defines the concept of http-URI, which is an absolute URI, but for the purposes of proxy request, the RFC uses the more general absolute-URI - because schemes other than http are allowed (including https).

snoyberg:
Thank you for the explanations, this makes sense now, and the code looks good overall. I have some minor comments that I'll make inline, but the biggest one is: I'm not sure about the naming of this setting. That's one of the reasons I asked for examples of this functionality from other libraries, to see how they name it. Looking at the RFC, I still don't see a better idea for naming. I'm slightly tempted to recommend something in the direction of calling out:

* Skipping the usage of `CONNECT`
* Highlighting the plain-text aspect of the connection

Do you have any thoughts on this naming topic?

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -286,6 +287,16 @@ noProxy = ProxyOverride $ const $ return $ \req -> req { proxy = Nothing }
 useProxy :: Proxy -> ProxyOverride
 useProxy p = ProxyOverride $ const $ return $ \req -> req { proxy = Just p }
 
+-- | Send secure requests to the proxy in plain text rather than using CONNECT,
+-- regardless of the value in the @Request@.
+--
+-- Since 0.7.2
+useProxySendRequestMode :: ProxyOverride -> ProxyOverride
```
snoyberg, line 294: This type signature feels a bit weird, I would have thought it would take a `Proxy` instead of `ProxyOverride`, just like `useProxy` does. Is there a specific reason to take a `ProxyOverride`.

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -286,6 +287,16 @@ noProxy = ProxyOverride $ const $ return $ \req -> req { proxy = Nothing }
 useProxy :: Proxy -> ProxyOverride
 useProxy p = ProxyOverride $ const $ return $ \req -> req { proxy = Just p }
 
+-- | Send secure requests to the proxy in plain text rather than using CONNECT,
+-- regardless of the value in the @Request@.
+--
+-- Since 0.7.2
```
snoyberg, line 293: For new code, I'd prefer to use the new `@since` syntax (I realize the other code doesn't use that.

Can you also make a cabal file bump and ChangeLog entry about this?

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -350,6 +352,13 @@ addProxy :: S.ByteString -> Int -> Request -> Request
 addProxy hst prt req =
     req { proxy = Just $ Proxy hst prt }
 
+
+-- | Send secure requests to the proxy in plain text rather than using CONNECT.
+--
+-- Since 0.7.2
```
snoyberg, line 358: Please also modify to `@since` here

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -466,7 +475,8 @@ requestBuilder req Connection {..} = do
         | otherwise  = fromByteString "http://"
 
     requestHostname
-        | isJust (proxy req) && not (secure req)
+        | isJust (proxy req) && (
+            proxySecureMode req == ProxySecureSendRequest || not (secure req))
```
snoyberg, line 479: The logic here is a little difficult to follow, and it's not resilient to possible increases in the number of `ProxySecureMode` data constructors in the future. Can we instead use more explicit pattern matching to hopefully clarify the meaning here, as well as cause a warning if an extra data constructor is added?

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -216,7 +217,7 @@ connKey req =
             | secure req -> simple CKSecure
             | otherwise -> simple CKRaw
         Just p
-            | secure req -> CKProxy
+            | secure req && proxySecureMode req == ProxySecureConnect -> CKProxy
```
snoyberg, line 220: Similar request here about pattern matching to catch increases in data constructors. Some comment, or well named helper function, may also be helpful to clarify why we make a special secure connection only when both `secure` is true _and_ we're using the `Connect` mode.

http-client/Network/HTTP/Client/Manager.hs
A1kmm: The intention was that it modifies `useProxy` rather than replaces it (i.e. usage would be `useProxySendRequestMode (useProxy $ Proxy ...))`). But happy to make it so you use it instead of `useProxy` and supply the Proxy directly to it.

A1kmm:
> ... tempted to recommend something in the direction of calling out:
> 
>     * Skipping the usage of `CONNECT`
> 
>     * Highlighting the plain-text aspect of the connection
> 
> 
> Do you have any thoughts on this naming topic?

Naming is hard and I don't think there is going to be a perfect name. But I think `proxyPlainTextNoCONNECTMode`, while a mouthful, would be intuitive enough about what it does.

snoyberg:
How about `SecureWithConnect` and `SecureWithoutConnect` as the general nomenclature? I think that gives enough context to reasonably understand the docs and at the very least guess at what the behavior may be.

A1kmm:
I have now pushed another commit to update it to do the renaming and address your other comments.

http-client/Network/HTTP/Client/Headers.hs
```diff
@@ -12,9 +12,7 @@ import           Control.Monad
 import qualified Data.ByteString                as S
 import qualified Data.ByteString.Char8          as S8
 import qualified Data.CaseInsensitive           as CI
-import           Data.Char (ord)
 import           Data.Maybe (mapMaybe)
-import           Data.Monoid
```
snoyberg, line 17: It looks like this change broke the GHC 8.2 builds.

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -210,21 +211,23 @@ getConn req m
...
-            | secure req -> CKProxy
-                (proxyHost p)
-                (proxyPort p)
-                (lookup "Proxy-Authorization" (requestHeaders req))
-                (host req)
-                (port req)
-            | otherwise -> CKRaw Nothing (proxyHost p) (proxyPort p)
-  where
-    simple con = con (hostAddress req) (host req) (port req)
+connKey req@Request { proxy = Nothing, secure = False } =
```
snoyberg, line 214: This looks great!

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -350,6 +352,13 @@ addProxy :: S.ByteString -> Int -> Request -> Request
 addProxy hst prt req =
     req { proxy = Just $ Proxy hst prt }
 
+
+-- | Send secure requests to the proxy in plain text rather than using CONNECT.
+--
+-- @since 0.7.2
+addProxyPlainTextNoCONNECTMode :: Request -> Request
```
snoyberg, line 359: I think this naming should match `useProxySecureWithoutConnect `

snoyberg (changes requested):
Looking good, just a few minor changes

http-client/Network/HTTP/Client/Request.hs
A1kmm: Sorry, I missed that one from my git add somehow... I've pushed the changes I already had locally now, and also changed Headers.hs to re-add `Data.Monoid` (I was hoping for a clean build with `-Wall -Werror` but it seems that will break compatibility).

snoyberg (approved):
Thanks!

http-client/Network/HTTP/Client/Manager.hs
mwotton: does this mean that there's no way to rely on the proxy set in the environment, but to also let it know that it should send requests in the clear?

PR #445, opened 2020-10-08
http-client-openssl: reasonable defaults

amesgen:
Moral successor to #106. Hostname validation (using OpenSSL's built-in support) is used. New dependency on [HsOpenSSL-x509-system](https://hackage.haskell.org/package/HsOpenSSL-x509-system) from @redneb.

 - The default configuration is not set in stone.
 - The negative SSL tests could be shared with http-client-tls (not sure if worth it).

### DIFF
--- a/.azure/azure-linux-template.yml
+++ b/.azure/azure-linux-template.yml
@@ -12,7 +12,7 @@ jobs:
       GHC 8.6:
         ARGS: "--resolver lts-14"
       GHC 8.8:
-        ARGS: "--resolver lts-15"
+        ARGS: "--resolver lts-16"
       nightly:
         ARGS: "--resolver nightly"
   steps:
--- a/.azure/azure-osx-template.yml
+++ b/.azure/azure-osx-template.yml
@@ -12,7 +12,7 @@ jobs:
       GHC 8.6:
         ARGS: "--resolver lts-14"
       GHC 8.8:
-        ARGS: "--resolver lts-15"
+        ARGS: "--resolver lts-16"
       nightly:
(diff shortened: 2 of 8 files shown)


amesgen:
CI fails as there now is a lower bound `HsOpenSSL >= 0.11.4.20`, and `0.11.4.20` is not yet present on Stackage.

One solution for this would be to not provide the defaults if `HsOpenSSL < 0.11.4.20`, as hostname validation is something every reasonable default should provide.

amesgen:
In [this commit](https://github.com/amesgen/http-client/commit/2d1be8460269cd81e4587c96df5e8dd42b5ee38a), I added an auto flag for the TLS defaults. I tested it with

```
cabal test --constraint='HsOpenSSL >= 0.11.4.20'
# and
cabal test --constraint='HsOpenSSL < 0.11.4.20'
```

amesgen:
There are multiple ways to get CI on green:

 1. Don't test `http-client-openssl` in CI.
 2. Modify CI to include appropriate versions of `HsOpenSSL` and `HsOpenSSL-x509-system` as `extra-deps`.
 3. Use cabal-install for testing `http-client-openssl`.

I went with 2. for now, but I am happy to change it. Any opinions?

alexbiehl (approved):
Thanks, I think this is looking great!

snoyberg:
LGTM. Can you include a changelog entry and version bump?

amesgen:
I squashed the three CI-related commits, but feel free to squash this entire PR.

snoyberg (approved):
Looks great, thank you!

PR #447, opened 2020-11-17
http-client-openssl: use hostAddress from Request

vshabanov:
Hi, `http-client-openssl` currently ignores `hostAddress` from `Request` making it unsuitable for use with external DNS resolver.

This PR fixes it by establishing HTTPS connections via the same functions that are used for plain HTTP connections.

Since it adds new function to `Network.HTTP.Client.Connection` the version of `http-client` should be bumped (and `http-client` dependency version in `http-client-openssl` must be bumped as well).

### DIFF
--- a/http-client-openssl/ChangeLog.md
+++ b/http-client-openssl/ChangeLog.md
@@ -1,3 +1,7 @@
+## 0.3.3
+
+* Use `hostAddress` from `Request`.
+
 ## 0.3.2.0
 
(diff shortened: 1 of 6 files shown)


snoyberg (changes requested):
LGTM. Please add in the version bumps, version bounds, and ChangeLog entries

vshabanov:
Added

snoyberg (approved):
Thanks!

PR #453, opened 2021-01-13
use base64-bytestring instead of memory

amesgen:
Only a tiny piece of `memory` was used, and we also get rid of the transitive dependency on `basement`.

Can add version bump and changelog if desired.


snoyberg:
Can you add a minor version bump and changelog entry?

snoyberg:
Thanks!

PR #454, opened 2021-02-01
Fix leaking sockets

### LINKED ISSUE #374: RequestTimeout setting results in `getProtocolByName: does not exist (no such protocol name: tcp))`
When sending several requests using `mapM_ (parseRequest ...)` to different destinations, I get many errors of the form
```
ghc: HttpExceptionRequest Request {
  host                 = "127.0.0.1"
  port                 = 45733
  secure               = True
  requestHeaders       = []
  path                 = "/"
  queryString          = ""
  method               = "POST"
  rawBody              = False
  redirectCount        = 10
}
 (InternalException (HostCannotConnect "127.0.0.1" [Network.Socket.socket: resource exhausted (Too many open files)]))
ghc: HttpExceptionRequest Request {
  host                 = "127.0.0.1"
  port                 = 40299
  secure               = True
  requestHeaders       = []
  path                 = "/"
  queryString          = ""
  method               = "POST"
  rawBody              = False
  redirectCount        = 10
}
 (ConnectionFailure Network.BSD.getProtocolByName: does not exist (no such protocol name: tcp))
```

This sounds like #307, but I'm only sending 25 requests, which hardly seems like DOS. Furthermore, I noticed that if I set  the Manager's managerResponseTimeout=responseTimeoutNone, I do not get any errors (everything works fine). After inspecting and playing with `getConnectionWrapper` in Network.HTTP.Client.Core, I see absolutely no reason why changing this setting affects this error. Yet it does, every single time.

There are several questions:
1. Why does changing the timeout setting affect whether I can call getProtocolbyName/run out of file descriptors?
2. I'm apparently getting *multiple* errors for each request I send (about 20 errors for each request, most of which are identical), which makes weeding through errors much harder. Why is that happening?
3. What is the root cause here?

swamp-agr:
Hi @snoyberg!

This PR is intended to close the issue with leaking sockets in case where `managerResponseTimeout` is relatively small. With outgoing constant 1K rps application becomes stuck after 4-17 hours uptime. Force reboot required every 4 hours. Situation is pretty similar to #374. Thus, this PR will close #374.

## Details

```haskell
withResponse req man f = bracket (responseOpen req man) responseClose f
```

- Timeout exception had been thrown  in "acquire" function of `bracket` (i.e. in `responseOpen`, when connection is opened). 
- Thus, connection could not be closed via `bracket`.
- It seems that `reaper` introduced in `Data.KeyedPool` did not manage such connections well. 
- It is also quite hard to get extra information from `KeyedPool` for further debugging purposes and produce the explicit test for such case. 
- Meantime, as it is currently implemented inside `httpRaw'` some exceptions are indeed handled well:

```haskell
    ...
    case ex of
        -- Connection was reused, and might have been closed. Try again
        Left e | managedReused mconn && mRetryableException m e -> do
            managedRelease mconn DontReuse
            httpRaw' req m
        -- Not reused, or a non-retry, so this is a real exception
        Left e -> throwIO e
        -- Everything went ok, so the connection is good. If any exceptions get
        -- thrown in the response body, just throw them as normal.
    ...
```

- In order to resolve the issue, separate case added:
```haskell
     ...
    case ex of
        -- Connection was reused, and might have been closed. Try again
        Left e | managedReused mconn && mRetryableException m e -> do
            managedRelease mconn DontReuse
            httpRaw' req m
        -- Connection timed out and might have been closed.
               | mTimeoutException m e -> do
            managedRelease mconn DontReuse
            throwIO e
        -- Not reused, or a non-retry, so this is a real exception
        Left e -> throwIO e
        -- Everything went ok, so the connection is good. If any exceptions get
        -- thrown in the response body, just throw them as normal.
    ...
```

- `mTimeoutException` added as `mRetryableException`, i.e. in a flexible way. Default implementation was also provided:

```haskell
    ...
    , managerTimeoutException = \e ->
        case fromException e of
          Just (HttpExceptionRequest _ ConnectionTimeout) -> True
          _                                               -> False
    ...
```

- It could be disabled via something like `const False` to fallback to current upstream.
- Haddock for `Manager` contains assumption about next package version (`@since 0.7.5`). I could remove it if you ask.
- ~I have `stylish-haskell` enabled by default. Thus, there are also some formatting changed. If formatting is not acceptable, I could proceed with reverting it.~

I am looking forward to bring the fix to upstream.

Best Regards,
@swamp-agr

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -126,31 +126,22 @@ httpRaw' req0 m = do
             Nothing -> return (req, res)
   where
     getConnectionWrapper mtimeout f =
-        bracketOnError getConnWithTimeout releaseConnMaybe checkConnDuration
-      where
-        -- If applicable, wait for timeout period to acquire the connection from the pool.
-        getConnWithTimeout = case mtimeout of
+        case mtimeout of
             Nothing -> fmap ((,) Nothing) f
             Just timeout' -> do
                 before <- getCurrentTime
                 mres <- timeout timeout' f
                 case mres of
-                  Nothing -> throwHttp ConnectionTimeout
-                  Just res -> return (Just (before, timeout'), res)
-
-        -- In case of errors in 'checkConnDuration' release connection.
-        releaseConnMaybe (_time, mconn) = managedRelease mconn DontReuse
-
-        -- Check connection duration and raise ConnectionTimeout if it takes too long.
-        checkConnDuration (mtime, mconn) = case mtime of
-            Nothing -> return (Nothing, mconn)
-            Just (before, timeout') -> do
-                now <- getCurrentTime
-                let timeSpentMicro = diffUTCTime now before * 1000000
-                    remainingTime = round $ fromIntegral timeout' - timeSpentMicro
-                if remainingTime <= 0
-                    then throwHttp ConnectionTimeout
-                    else return (Just remainingTime, mconn)
+                     Nothing -> throwHttp ConnectionTimeout
+                     Just mConn -> do
+                         now <- getCurrentTime
+                         let timeSpentMicro = diffUTCTime now before * 1000000
+                             remainingTime = round $ fromIntegral timeout' - timeSpentMicro
+                         if remainingTime <= 0
+                             then do
+                                 managedRelease mConn DontReuse
+                                 throwHttp ConnectionTimeout
+                             else return (Just remainingTime, mConn)
 
     responseTimeout' req =
         case responseTimeout req of
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -110,6 +110,10 @@ httpRaw' req0 m = do
         Left e | managedReused mconn && mRetryableException m e -> do
             managedRelease mconn DontReuse
             httpRaw' req m
+        -- Connection timed out and might have been closed.
+               | mTimeoutException m e -> do
+            managedRelease mconn DontReuse
+            throwIO e
         -- Not reused, or a non-retry, so this is a real exception
         Left e -> throwIO e
         -- Everything went ok, so the connection is good. If any exceptions get
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -110,12 +110,12 @@ httpRaw' req0 m = do
         Left e | managedReused mconn && mRetryableException m e -> do
             managedRelease mconn DontReuse
             httpRaw' req m
-        -- Connection timed out and might have been closed.
-               | mTimeoutException m e -> do
-            managedRelease mconn DontReuse
-            throwIO e
         -- Not reused, or a non-retry, so this is a real exception
-        Left e -> throwIO e
+        Left e -> do
+          -- Explicitly release connection for all real exceptions:
+          -- https://github.com/snoyberg/http-client/pull/454
+          managedRelease mconn DontReuse
+          throwIO e
         -- Everything went ok, so the connection is good. If any exceptions get
         -- thrown in the response body, just throw them as normal.
         Right res -> case cookieJar req' of
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -126,20 +126,31 @@ httpRaw' req0 m = do
             Nothing -> return (req, res)
   where
     getConnectionWrapper mtimeout f =
-        case mtimeout of
+        bracketOnError getConnWithTimeout releaseConnMaybe checkConnDuration
+      where
+        -- If applicable, wait for timeout period to acquire the connection from the pool.
+        getConnWithTimeout = case mtimeout of
             Nothing -> fmap ((,) Nothing) f
             Just timeout' -> do
                 before <- getCurrentTime
                 mres <- timeout timeout' f
                 case mres of
-                    Nothing -> throwHttp ConnectionTimeout
-                    Just res -> do
-                        now <- getCurrentTime
-                        let timeSpentMicro = diffUTCTime now before * 1000000
-                            remainingTime = round $ fromIntegral timeout' - timeSpentMicro
-                        if remainingTime <= 0
-                            then throwHttp ConnectionTimeout
-                            else return (Just remainingTime, res)
+                  Nothing -> throwHttp ConnectionTimeout
+                  Just res -> return (Just (before, timeout'), res)
+
+        -- In case of errors in 'checkConnDuration' release connection.
+        releaseConnMaybe (_time, mconn) = managedRelease mconn DontReuse
+
+        -- Check connection duration and raise ConnectionTimeout if it takes too long.
+        checkConnDuration (mtime, mconn) = case mtime of
+            Nothing -> return (Nothing, mconn)
+            Just (before, timeout') -> do
+                now <- getCurrentTime
+                let timeSpentMicro = diffUTCTime now before * 1000000
+                    remainingTime = round $ fromIntegral timeout' - timeSpentMicro
+                if remainingTime <= 0
+                    then throwHttp ConnectionTimeout
+                    else return (Just remainingTime, mconn)
 
     responseTimeout' req =
         case responseTimeout req of
--- a/http-client/Network/HTTP/Client/Manager.hs
+++ b/http-client/Network/HTTP/Client/Manager.hs
@@ -205,13 +205,7 @@ getConn req m
     -- Stop Mac OS X from getting high:
     -- https://github.com/snoyberg/http-client/issues/40#issuecomment-39117909
     | S8.null h = throwHttp $ InvalidDestinationHost h
-    | otherwise = 
-        -- Release connection in case of connection timeout:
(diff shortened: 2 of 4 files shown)


http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -20,22 +20,23 @@ module Network.HTTP.Client.Manager
...
+import qualified Network.Socket                 as NS
 
-import Network.HTTP.Types (status200)
-import Network.HTTP.Client.Types
-import Network.HTTP.Client.Connection
-import Network.HTTP.Client.Headers (parseStatusHeaders)
-import Network.HTTP.Proxy
-import Data.KeyedPool
-import Data.Maybe (isJust)
+import           Data.KeyedPool
```
snoyberg, line 33: I'd prefer to receive a PR without unnecessary code formatting changes. Can you resubmit with just the relevant changes?

snoyberg (changes requested):
I've started reviewing, but I'd rather look at a clean diff before adding any comments, as I may be missing things.

http-client/Network/HTTP/Client/Manager.hs
swamp-agr: Good! Please expect changes soon.

swamp-agr: Formatting changes undone, commits squashed. Now the diff contains only relevant changes.

snoyberg:
Thanks for updating. I'm not sure I fully understand the motivation behind the change here though. It seems like the problem is that, if an exception is thrown from a reused connection, we need to ensure that it is released. My question is: what makes timeout exceptions special? Shouldn't we simply call `managedRelease mconn DontReuse` for all types of exceptions?

Yuras:
@swamp-agr Let me describe how it works right now. `KeyedPool` doesn't handle allocated connections at all, `reaper` is for idle connections (waiting to be reused).

When new connection is allocated, weak finalizer is attached to it to deallocate it in case it leaks, but it doesn't guarantee prompt deallocation, see [here](https://github.com/snoyberg/http-client/blob/897b307c0bfad1ed63554f9838aec604cf41a4e4/http-client/Data/KeyedPool.hs#L213).

Indeed `http-client` doesn't even try to guarantee prompt connection deallocation, it relies mostly on finalizers.

The issue in question occurs after 4-7 hours of correct work, so I assume there where multiple major GCs during this time, so finalizers had enough time to be executed. I see three possibilities here:
- `KeyedPool` is buggy and doesn't attached finalizer is some cases, or
- Something keeps a reference to the connections (probably via `Response`), preventing finalizers from running, or
- Finalizers are broken in ghc, e.g. you are using older ghc (pre `8.6` IIRC) or they are still broken in newer ghc.

So my conclusion: it's likely that you are fixing a symptom instead of the core issue.

There is nothing wrong with fixing symptoms, and actually I never liked the fact that `http-client` relies on finalizers. Now the following question comes: do you want `http-client` to guarantee prompt deallocation? If you do, then the PR is way too simple for that, but if you don't, then it's probably OK.

And I agree with @snoyberg that request timeout is not special in any way. Consider for example the [connection timeout](https://github.com/snoyberg/http-client/blob/897b307c0bfad1ed63554f9838aec604cf41a4e4/http-client/Network/HTTP/Client/Core.hs#L131), with high probability it will trigger the same issue. Not to mention IO errors or async exceptions.

swamp-agr:
@Yuras, thank you for detailed explanation.

@snoyberg, I want to provide wide fix for deallocation of the `httpRaw'` function. 
There are at least two entry points of affected deallocation: 
- [`getConnectionWrapper`](https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Core.hs#L95).
- And [response exceptions](https://github.com/snoyberg/http-client/blob/master/http-client/Network/HTTP/Client/Core.hs#L108).

In this case the next steps for me would be:

- close connections before throwing exception in `getConnectionWrapper`.
- close connections for all exceptional results of `getResponse` (`Left e -> throwIO e`).
- remove `Timeout` related settings and functions.

Please expect fix soon.

swamp-agr:
Changes that were described in comment above implemented now.

http-client/Network/HTTP/Client/Manager.hs
```diff
@@ -205,7 +205,13 @@ getConn req m
     -- Stop Mac OS X from getting high:
     -- https://github.com/snoyberg/http-client/issues/40#issuecomment-39117909
     | S8.null h = throwHttp $ InvalidDestinationHost h
-    | otherwise = takeKeyedPool (mConns m) connkey
+    | otherwise = 
+        -- Release connection in case of connection timeout:
+        -- https://github.com/snoyberg/http-client/pull/454
+        bracketOnError
```
snoyberg, line 211: This call is a nop. `bracketOnError` will only trigger the cleanup if the in-between action throws an exception. If the allocation action throws an exception, the finalizer will never be run.

swamp-agr: I thought about that. On the other hand, in case of exception `Managed Connection` will not be created and it will not be an issue. Is it correct or I am missing something?

snoyberg: That's correct. I'm not sure what problem you're trying to prevent here.

swamp-agr: - I was concerned about `timeout timeout' (getConn req m)` call in `getConnectionWrapper`. 
- I thought that in some tricky case connection could be acquired but dropped via `timeout`. In this case its future will be uncertain and I tried to prevent it.
- If `Managed Connection` created before timeout, it definitely would be returned. Otherwise, it would not be created at all. 
- And if it would not be created, then `ConnectionTimeout` exception could be "safely" thrown from `getConnectionWrapper`.

Now I see, there is nothing to worry about. I will completely restore `getConn`. 

Thank you for your comment.

swamp-agr: Fixed.

snoyberg (changes requested):
LGTM! Only last request is a changelog entry.

@Yuras did you have any additional comments before I merge this one in?

swamp-agr:
@snoyberg changelog updated accordingly (I assume that next version will be `0.7.5`).

swamp-agr:
After quick chat with @Yuras he pointed me out on `getConnectionWrapper`:

```haskell
  where
    getConnectionWrapper mtimeout f =
        case mtimeout of
            Nothing -> fmap ((,) Nothing) f
            Just timeout' -> do
                before <- getCurrentTime
                mres <- timeout timeout' f
                case mres of
                    Nothing -> throwHttp ConnectionTimeout
                    Just res -> do
                        now <- getCurrentTime
                        let timeSpentMicro = diffUTCTime now before * 1000000
                            remainingTime = round $ fromIntegral timeout' - timeSpentMicro
                        if remainingTime <= 0
                            then throwHttp ConnectionTimeout -- <-- In this case connection has already present, it would leak
                            else return (Just remainingTime, res)
```

There is also possibility of asynchronous exceptions. 
Thus, his recommendation is to apply `bracketOnError` around `getConnectionWrapper`.

swamp-agr:
Suggestion implemented. 
@Yuras, @snoyberg please review.

CI errors from appveyour could be ignored, GitHub webhooks are degraded right now:

telegram-cloud-photo-size-2-5249229594207301805-y

http-client/Network/HTTP/Client/Core.hs
```diff
@@ -122,20 +126,31 @@ httpRaw' req0 m = do
...
     getConnectionWrapper mtimeout f =
-        case mtimeout of
+        bracketOnError getConnWithTimeout releaseConnMaybe checkConnDuration
+      where
+        -- If applicable, wait for timeout period to acquire the connection from the pool.
+        getConnWithTimeout = case mtimeout of
             Nothing -> fmap ((,) Nothing) f
             Just timeout' -> do
                 before <- getCurrentTime
                 mres <- timeout timeout' f
```
snoyberg, line 136: This is a problem. `timeout` is now running in a masked state, which we shouldn't be doing.

swamp-agr: Should `timeout` function be wrapped in `interruptible` explicitly?

Yuras: @snoyberg Why is it a problem? `getConn` is interruptible, so `timeout` will be able to interrupt it. And in any case, `getConn` basically just calls `takeKeyedPool`, with is `mask_`'ed anyway.

snoyberg: I don't want to rely on the current behavior of `bracketOnError` not masking in the acquire clause, that may change in the future. It's fragile to write code that way IMO. It seems like the primary thing we're trying to achieve is ensuring that `releaseConnMaybe` is called in the case of `remainingTime <= 0`. Why not add that in directly before the `throwHttp` call, or add an `onException`?

swamp-agr: Restored previous version and handled directly. Fixed.

snoyberg (approved):
This looks good. Let’s start with this change, and we can continue with other improvements as necessary

swamp-agr:
@snoyberg @Yuras Thank you for your support!

Even with early version of the fix situation changed.
According to `netstat` and application logs, bottleneck shifted from `http-client` side to `warp` side.

Please expect PR in `warp` on the next week.

PR #456, opened 2021-02-20
Add setRequestResponseTimeout convenience function

MaxGabriel:
I like to use the convenience functions, and it's been annoying this one is missing.

I haven't tested this in practice, but this seems pretty straightforward too?

PR #457, opened 2021-02-20
Add bearer auth functions

### LINKED ISSUE #455: Convenience function for setting bearer authentication
Currently there is a helper function for `setRequestBasicAuth`. Would it be desired to have a similar function for adding Bearer auth? This would look something like the following:

`Authorization: Bearer <base64encodedvalue>`

Examples of APIs that use bearer authentication are [Front](https://dev.frontapp.com/docs/core-api-getting-started) and [Stripe](https://stripe.com/docs/api/authentication) (Stripe accepts basic and bearer)

MaxGabriel:
Closes #455

Notably different from #455, I researched some more and bearer auth does _not_ imply base64 encoding the way basic auth does.

Also fwiw, many tools (eg curl) don't seem to have bearer auth helper functions. That said, still seems nice to have to me?

I tested this addition on a real API as well as the added httpbin tests

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 0.7.6
 
-* Add `applyBearerAuth` function []()
+* Add `applyBearerAuth` function [#457](https://github.com/snoyberg/http-client/pull/457/files)
 
 ## 0.7.5
 
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## 0.7.6
+
+* Add `applyBearerAuth` function []()
+
 ## 0.7.5
 
 * Force closing connections in case of exceptions throwing [#454](https://github.com/snoyberg/http-client/pull/454).
--- a/http-conduit/ChangeLog.md
+++ b/http-conduit/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 2.3.8
 
-* Adds `setRequestBearerAuth` convenience function. Note that this is only available for `http-client` versions 0.7.6 or greater. []()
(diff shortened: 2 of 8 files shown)


http-conduit/Network/HTTP/Simple.hs
```diff
@@ -76,6 +76,9 @@ module Network.HTTP.Simple
     , H.setRequestIgnoreStatus
     , H.setRequestCheckStatus
     , setRequestBasicAuth
+#if MIN_VERSION_http_client(0,7,6)
```
MaxGabriel, line 79: Using CPP for backwards compat with old http-clients matches the existing usage for `shouldStripHeaderOnRedirect`

http-client/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## 0.7.6
+
+* Add `applyBearerAuth` function []()
```
MaxGabriel, line 5:
```suggestion
* Add `applyBearerAuth` function [#457](https://github.com/snoyberg/http-client/pull/457/files)
```

http-conduit/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # ChangeLog for http-conduit
 
+## 2.3.8
+
+* Adds `setRequestBearerAuth` convenience function. Note that this is only available for `http-client` versions 0.7.6 or greater. []()
```
MaxGabriel, line 5:
```suggestion
* Adds `setRequestBearerAuth` convenience function. Note that this is only available for `http-client` versions 0.7.6 or greater. [#457](https://github.com/snoyberg/http-client/pull/457/files)
```

snoyberg (approved):
LGTM, thanks! I'll hold off on releasing in case you want to bundle #456 in here too.

PR #460, opened 2021-04-12
Allow secure cookies for localhost

### LINKED ISSUE #459: Allow secure cookies for localhost
[Firefox 75](https://bugzilla.mozilla.org/show_bug.cgi?id=1618113) and [Chrome 89](https://bugs.chromium.org/p/chromium/issues/detail?id=1056543) now allow "Secure" cookies to be sent over plain http when set by localhost.

This is fully spec compliant, see both linked issues for details.

Should we consider a similar change for `http-client`?

It would require a little additional logic [here](https://github.com/snoyberg/http-client/blob/0a4056efb014766a3d6f099d0878c58d07144251/http-client/Network/HTTP/Client/Cookies.hs#L146).

The benefit is, in a lot of cases, you'd be able to just `{ setCookieSecure = True }` like one probably should, without having to disable it for testing and development.

This also affects downstream libraries like `wreq`.

mulderr:
Closes #459

Up for review. Still testing.

### DIFF
--- a/http-client/Network/HTTP/Client/Cookies.hs
+++ b/http-client/Network/HTTP/Client/Cookies.hs
@@ -14,6 +14,7 @@ module Network.HTTP.Client.Cookies
     , removeExistingCookieFromCookieJar
     , domainMatches
     , isIpAddress
+    , isPotentiallyTrustworthyOrigin
     , defaultPath
     ) where
 
@@ -29,6 +30,8 @@ import Blaze.ByteString.Builder
 import qualified Network.PublicSuffixList.Lookup as PSL
 import Data.Text.Encoding (decodeUtf8With)
 import Data.Text.Encoding.Error (lenientDecode)
+import qualified Data.IP as IP
+import Text.Read (readMaybe)
 
 import Network.HTTP.Client.Types as Req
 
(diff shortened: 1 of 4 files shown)


http-client/Network/HTTP/Client/Cookies.hs
```diff
@@ -111,6 +114,37 @@ rejectPublicSuffixes = True
...
+  | isLoopbackAddr6 = True    -- step 4, part 2
+  | isLoopbackHostname = True -- step 5
+  | otherwise = False
+  where isLoopbackHostname =
+               host == "localhost"
+            || host == "localhost."
+            || BS.isSuffixOf ".localhost" host
+            || BS.isSuffixOf ".localhost." host
+        isLoopbackAddr4 =
+          fmap (head . IP.fromIPv4) (readMaybe (S8.unpack host)) == Just 127
```
snoyberg, line 138: I'm uncomfortable about usage of a partial function here, `take 1` would be better

snoyberg (changes requested):

mulderr:
I've noticed `Read` instances for [`IPv4`](http://hackage.haskell.org/package/iproute-1.7.11/docs/src/Data.IP.Addr.html#ip4) and [`IPv6`](http://hackage.haskell.org/package/iproute-1.7.11/docs/src/Data.IP.Addr.html#ip6) skip leading whitespace. Not sure if this creates any real issues but just to be safe I'd rather refactor to use the parsers directly and be as strict as possible.

mulderr:
On the other hand it's already parsed by network-uri and it's an internal function therefore I guess it's fine. Let me now if you'd prefer a stricter version.

snoyberg:
I'm OK with this as-is. Have you finished your testing? Is this ready for final review?

mulderr:
Yep! Looks good but I'm of course happy to fix any remaining issues.

snoyberg:
Only other thing is some bookkeeping. Please:

* Perform a minor version bump in the .cabal file
* Add a ChangeLog entry

snoyberg (approved):
Thank you!

PR #464, opened 2021-04-20
Add original Request to the Response type.

miguelclean:
Another proposal to get the original Request within `managerModifyResponse`.

Example usage:
```
main :: IO ()
main = do
  manager <- newManager defaultManagerSettings { managerModifyResponse = logRsp }
  request <- parseRequest "http://httpbin.org/get"
  response <- httpLbs request manager
  print $ responseBody response
  where
    logRsp :: Response a -> IO (Response a)
    logRsp response = do
      putStrLn $  "Response to request: " <> show (path $ responseOriginalRequest response) <> " = "
                                          <> show (responseStatus response)
      pure response
```

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## 0.7.8
+
+* Include the original `Request` in the `Response`. Expose it via `getOriginalRequest`.
+
 ## 0.7.7
 
 * Allow secure cookies for localhost without HTTPS [#460](https://github.com/snoyberg/http-client/pull/460)
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -181,6 +181,7 @@ module Network.HTTP.Client
     , responseHeaders
     , responseBody
     , responseCookieJar
+    , getOriginalRequest
     , throwErrorStatusCodes
       -- ** Response body
     , BodyReader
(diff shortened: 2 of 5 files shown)


snoyberg:
I think I'm on board with this one, and I like the naming. It strongly implies that this is the request _before_ any redirects are followed. Some considerations:

* Before merging, this should include a minor version bump and changelog entry (I'm guessing you were already planning that based on the `@since` waiting to be filled in)
* To avoid confusion with the request body, perhaps replacing the request body with an empty one would be a Good Thing, with that condition documented
* Instead of directly exposing the accessors, I'd rather expose a function `getOriginalRequest` or similar, so that it can be used for setting values. I realize that's _not_ how the rest of the fields work, but it's how I wish I'd written them

miguelclean:
Thank you for the positive response. I just updated this pull request accordingly, however I'm still not sure if it meets the expectations.

In particular I was not sure where to place the `getOriginalRequest` function. Beside that, now its documenation seems to duplicate the (non-public) documentation of the `responseOriginalRequest` field itself.

snoyberg:
It actually looks perfect to me. Only final request I'd make is to add a comment on `responseOriginalRequest` explaining that it's _intentionally_ not exported, and that we export `getOriginalRequest` instead.

miguelclean:
Glad to hear that. I updated the comment on `responseOriginalRequest`.

snoyberg (approved):
Looks great, thank you!

avieth:
I know this is already released so it's maybe a fool's errand, but I want to propose this alternative implementation which doesn't require having a `Request` in order to make a `Response` value https://github.com/snoyberg/http-client/pull/488

Issue #469, opened 2021-08-31
Default value of onRequestBodyException causes exceptions to be ignored when streaming body


ondrap:
The default onRequestBodyException is:

```haskell
 onRequestBodyException = \se ->
            case E.fromException se of
                Just (_ :: IOException) -> return ()
                Nothing -> throwIO se
```

The problem is that when using the streaming or chunked streaming body, if an `IOException` is thrown, it is ignored. The problem is that it means that if I'm e.g. streaming the body from a file (e.g. using amazonka-s3 and default conduit) and an `IOException` occurs (e.g. Permission denied), the whole request just hangs. 

I'm probably going to solve this by recasting the exception to something else (and writing my own file-reading conduit), as there seems to be no way to override this default in amazonka. However it seems to me a rather weird default, does it have any reason to be there?

snoyberg:
It's a weird case, but the description is in the Git commit: ae82e09c4658babf0aca7edd98669f0b941beeae. For the case you describe, I would recommend opening up the file handle _before_ starting the `Request`, if possible.

Perhaps we could be more specific in which `IOException`s we ignore here.

ondrap:
My plan is to ultimately stream directly from an FTP connection, so I don't think that would help. However, it seems to me it would be better to solve it the 'reverse' way I did it in my code - what about catching the IO exception in the `writeBuilder`, rethrowing it as some custom exception, catching that, if it is an early close, then ignore, otherwise rethrow it as a normal exception again?

snoyberg:
I hear that, and in a from-scratch implementation that would probably be a better approach. However, in this case, I'm concerned about such a subtle change to behavior that's been around for a very long time in the library.

snoyberg:
How's this: would you be up for putting out a message on the yesodweb.com blog proposing the change (maybe even just a link to a proposed PR) and we'll see if anyone has complaints? That seems like a good way to get publicity on this before making a change.

ondrap:
Sure, that would be great. Should I try to prepare the PR?

snoyberg:
That would be great!

PR #470, opened 2021-09-07
Route Popper exceptions around onRequestBodyException handler.

ondrap:
Exceptions thrown in the Popper (streaming request body) are routed around the `onRequestBodyException` handler. 

Previously if the Popper has thrown an exception, the default handler would silently discard that. Even installing custom handler made it impossible to distinguish between these kind of exceptions, resulting in timed out requests instead of immediate failure.

See #469

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog for http-client
 
+## 0.7.9
+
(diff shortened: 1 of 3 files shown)


snoyberg:
Looks great! Can you add a minor version bump and a ChangeLog entry?

ondrap:
Done.

snoyberg (approved):
Thanks!

PR #471, opened 2021-09-13
Document Network.HTTP.Client.OpenSSL

robx:
I had to read the source to understand what's going on with `makeDefaultContext` in particular. I can't say I really understand the reasoning behind the API design, so it's quite possible this could be improved, but at least it should be apparent now that `newOpenSSLManager` isn't just `newManager $ openSSLManagerSettings $ defaultMakeContext defaultOpenSSLSettings` as one might expect.


snoyberg (approved):
Looks great, thanks!

PR #472, opened 2021-12-20
Asynchronous search for successful IP

lolepezy:
This is to introduce a minimal (and not complete) implementation of RFC 6555 and RFC 8305. 

- Instead of doing manual asynchronous DNS lookup for AAAA and A records, it relies on getAddrInfo.
- Available IP addresses are chosen concurrently.

The reason for the change is to avoid long timeouts when a domain has AAAA DNS record but no IPv6 connectivity, which is a relatively common misconfiguration.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -188,28 +188,32 @@ openSocket tweakSocket addr =
             tweakSocket sock
             NS.connect sock (NS.addrAddress addr)
             return sock)
-        
+
 -- Pick up an IP using an approximation of the happy-eyeballs algorithm:
 -- https://datatracker.ietf.org/doc/html/rfc8305
 -- 
 firstSuccessful :: [NS.AddrInfo] -> (NS.AddrInfo -> IO a) -> IO a
 firstSuccessful []        _  = error "getAddrInfo returned empty list"
-firstSuccessful addresses cb = do             
+firstSuccessful addresses cb = do
     result <- newEmptyMVar
-    either E.throwIO pure =<< 
-        withAsync (tryAddresses result) 
-            (\_ -> takeMVar result)    
+    either E.throwIO pure =<<
+        withAsync (tryAddresses result)
+            (\_ -> takeMVar result)
   where
     -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
-    connectionAttemptDelay = 250 * 1000    
+    connectionAttemptDelay = 250 * 1000
 
-    tryAddresses lastResult = do         
-        z <- forConcurrently (zip addresses [0..]) $ \(addr, n) -> do
-                when (n > 0) $ threadDelay $ n * connectionAttemptDelay        
-                r :: Either E.IOException a <- E.try $ cb addr            
-                for_ r $ \_ -> tryPutMVar lastResult r
-                pure r
+    tryAddresses result = do
+        lastException <- take 1 . filter isLeft <$> go addresses []
+        for_ lastException $ tryPutMVar result
+      where
+        go []             asyncs = mapM wait asyncs
+        go (addr : addrs) asyncs =
+            withAsync (tryAddress addr) $ \a -> do
+                unless (null addrs) $ threadDelay connectionAttemptDelay
+                go addrs $ a : asyncs
 
-        let lastException = take 1 $ reverse $ filter isLeft z
-        for_ lastException $ tryPutMVar lastResult        
-
+        tryAddress addr = do
+            r :: Either E.IOException a <- E.try $ cb addr
+            for_ r $ \_ -> tryPutMVar result r
+            pure r
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -214,6 +214,6 @@ firstSuccessful addresses cb = do
                 go addrs $ a : asyncs
 
         tryAddress addr = do
-            r :: Either E.IOException a <- E.try $ cb addr
+            r :: Either E.IOException a <- E.try $! cb addr
             for_ r $ \_ -> tryPutMVar result r
             pure r
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -192,43 +192,25 @@ openSocket tweakSocket addr =
 -- 
 firstSuccessful :: [NS.AddrInfo] -> (NS.AddrInfo -> IO a) -> IO a
 firstSuccessful []        _  = error "getAddrInfo returned empty list"
-firstSuccessful addresses cb = do     
-    results <- newTVarIO []
-    let totalSize = length addresses
-    withAsync (tryAddresses addresses results) $ \_ -> 
-        waitForResult results totalSize   
-  where    
-    tryConnecting addr results = do 
-        r <- E.try $! cb addr                
-        atomically $ modifyTVar' results (r:)
-     
-    -- Try to connect to every address concurrently with a delay.
-    -- Whichever succeeds first wins.
-    tryAddresses adresses results = go adresses []
-      where
-        go []             asyncs = mapM_ waitCatch asyncs 
-        go (addr : addrs) asyncs = 
-            withAsync (tryConnecting addr results) $ \a -> do 
-                -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
-                let connectionAttemptDelay = 250 * 10000
-                threadDelay connectionAttemptDelay
-                go addrs (a : asyncs)
+firstSuccessful addresses cb = do             
+    lastResult <- newTVarIO Nothing
 
-    waitForResult results totalSize = 
-        atomically $ do 
-            rs <- readTVar results
-            case rs of
-                [] -> retry
-                z  ->
-                    case [ r | Right r <- z ] of
-                        []
-                            -- tried all of them, but there're no successful ones
-                            | length z == totalSize -> 
-                                throwSTM $ head [ e :: E.SomeException | Left e <- z ]
+    result <- withAsync (tryAddresses lastResult) $ \_ -> 
+                atomically $ maybe retry pure =<< readTVar lastResult                        
 
-                            -- there are still addresses to try
-                            | otherwise -> retry
+    either E.throwIO pure result    
+  where
+    -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
+    connectionAttemptDelay = 250 * 1000    
 
-                        -- at least one connection was successful
-                        as -> pure $ last as  
-                
+    tryAddresses lastResult = 
+          forConcurrently_ (zip addresses [0..]) $ \(addr, n) -> do
+            when (n > 0) $ threadDelay $ n * connectionAttemptDelay        
+            r :: Either E.IOException a <- E.try $ cb addr            
+            atomically $ do                   
+                previous <- readTVar lastResult 
+                case (previous, r) of 
+                    (Nothing,        _) -> writeTVar lastResult (Just r)
+                    (Just (Left  _), _) -> writeTVar lastResult (Just r)
+                    (Just (Right _), _) -> pure ()
+
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -187,7 +187,7 @@ openSocket tweakSocket addr =
             NS.connect sock (NS.addrAddress addr)
             return sock)
 
--- Resolve using an approximation of the happy-eyeballs algorithm:
+-- Pick up an IP using an approximation of the happy-eyeballs algorithm:
 -- https://datatracker.ietf.org/doc/html/rfc8305
 -- 
 firstSuccessful :: [NS.AddrInfo] -> (NS.AddrInfo -> IO a) -> IO a
@@ -222,7 +222,7 @@ firstSuccessful addresses cb = do
                 z  ->
                     case [ r | Right r <- z ] of
                         []
-                            -- tried all of them, but there's no successful ones
+                            -- tried all of them, but there're no successful ones
                             | length z == totalSize -> 
                                 throwSTM $ head [ e :: E.SomeException | Left e <- z ]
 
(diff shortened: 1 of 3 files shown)


http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -184,10 +187,58 @@ openSocket tweakSocket addr =
...
+      where
+        go [addr] asyncs = do 
+            tryConnecting addr results
+            waitForThem asyncs     
+
+        go (addr : addrs) asyncs = 
+            withAsync (tryConnecting addr results) $ \a -> do             
+                -- connection attempt delay 
+                -- https://datatracker.ietf.org/doc/html/rfc8305#section-5                
+                threadDelay $ 250 * 10000
```
snoyberg, line 218: I think the intend is 250 milliseconds, but that would be `250 * 1000`, not `250 * 10000`.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -184,10 +187,58 @@ openSocket tweakSocket addr =
...
+-- https://datatracker.ietf.org/doc/html/rfc8305
+-- 
 firstSuccessful :: [NS.AddrInfo] -> (NS.AddrInfo -> IO a) -> IO a
-firstSuccessful []     _  = error "getAddrInfo returned empty list"
-firstSuccessful (a:as) cb =
-    cb a `E.catch` \(e :: E.IOException) ->
-        case as of
-            [] -> E.throwIO e
-            _  -> firstSuccessful as cb
+firstSuccessful []        _  = error "getAddrInfo returned empty list"
```
snoyberg, line 194: I'm overall having trouble following the intended logic of this code. If I understand it correctly, I think it could be simplified by usage of `mapConcurrently_` and `race`, something along the lines of:

```haskell
firstFoundAddress <- newEmptyTMVarIO
result <- race (atomically $ takeTMVar firstFoundAddress) $ forConcurrently_ addresses $ \addr -> do
  r :: Either E.IOException _ <- try $ cb addr
  for_ r $ atomically . tryPutTMVar firstFoundAddress
either pure $ \() -> throwIO NoAddressFound
```

That's skipping over some of the details, but heads in the direction I was thinking of.

http-client/Network/HTTP/Client/Connection.hs
lolepezy: Indeed! Thanks, will fix it.

http-client/Network/HTTP/Client/Connection.hs
lolepezy: It is indeed quite cumbersome, the reasons for which were these

- support the delay between opening sockets (RFC prescribes that and it's just reasonable to not try to connect to all addresses at once)
- minimize the number of threads
- keep the original semantics of "if no success, throw the exception that happened the last", since I don't know what can break if changed (most likely, nothing).

So if you are okay with replacing original connection exceptions with `NoAddressFound` and don't mind creating all `asyncs` at once,  I would test something like 
```
    firstFoundAddress <- newEmptyTMVarIO
    result <- race (atomically $ takeTMVar firstFoundAddress) $ 
                forConcurrently_ (zip addresses [0..]) $ \(addr, n) -> do
        when (n > 0) $ threadDelay $ n * delay
        r :: Either E.IOException _ <- try $ cb addr
        for_ r $ atomically . tryPutTMVar firstFoundAddress
    either pure $ \() -> throwIO NoAddressFound
```

lolepezy: I pushed a version that doesn't break any test and is relatively idiomatic. It still creates more threads than possible, but for the most practical cases the difference should be acceptable.

snoyberg: I don't mind replacing the exception returned, though using another `TMVar` to capture "first exception raised" should work as well. I'm not sure how many threads we're expecting here. How many addresses would typically be returned?

One very subtle bug that may arise with the code I provided: it's possible that GHC's runtime will mistakenly think there's a deadlock going on. It's probably best to instead have the `TMVar` contain an `Either IOException _` and populate the field after the `forConcurrently_` call.

lolepezy: I managed to come up with a version that looks relatively idiomatic, should not deadlock and doesn't break tests on any OS anymore.

> How many addresses would typically be returned?

That is not so easy to say, I would guess that typical host would have 2, but since a lot of people deploy to clouds and use CDNs, they may have quite a few. Here are just some examples from my use case:
```
λ length <$> getAddrInfo (Just hints) (Just "rrdp.ripe.net") (Just "443")
12
λ length <$> getAddrInfo (Just hints) (Just "ripe.net") (Just "443")
2 
λ length <$> getAddrInfo (Just hints) (Just "cc.rg.net") (Just "443")
2
λ length <$> getAddrInfo (Just hints) (Just "rrdp.arin.net") (Just "443")
7
```
The one with a dozen of addresses is the CDN. I would assume that in majority of cases the very first address will be returned before others try to connect, so we'll have a bunch of asyncs sitting in a `threadDelay`. I can come up with something more ugly and efficient with `withAsync` and go-pattern, just to be sure we are not wasting threads.

snoyberg:
LGTM. Could you add a minor version bump and changelog entry?

lolepezy:
Yeap, done.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -183,11 +188,28 @@ openSocket tweakSocket addr =
...
+    either E.throwIO pure =<< 
+        withAsync (tryAddresses result) 
+            (\_ -> takeMVar result)    
+  where
+    -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
+    connectionAttemptDelay = 250 * 1000    
+
+    tryAddresses lastResult = do         
+        z <- forConcurrently (zip addresses [0..]) $ \(addr, n) -> do
+                when (n > 0) $ threadDelay $ n * connectionAttemptDelay
```
snoyberg, line 208: Just thinking this through a little bit more, the current behavior is:

* Fork N threads immediately
* N-1 of those threads sleep for 250us

What if instead:

* Fork two threads immediately
* First thread immediately calls `cb addr` on the first address
* Second thread sleeps for 250us and then forks the other N-1 threads

lolepezy: Would it make sense to apply this logic recursively?

This version works and I don't see what may be wrong with (except that it can be made better with `foldM` probably).

```
    tryAddresses lastResult = do        
        lastException <- take 1 . reverse . filter isLeft <$> go addresses
        for_ lastException $ tryPutMVar lastResult
      where
        go []             = pure []
        go (addr : addrs) = 
            uncurry (:) <$>
                concurrently 
                    (tryAddress addr) 
                    (do 
                        unless (null addrs) (threadDelay connectionAttemptDelay) 
                        go addrs)

        tryAddress addr = do
            r :: Either E.IOException a <- E.try $ cb addr
            for_ r $ \_ -> tryPutMVar lastResult r
            pure r
 ```

Comment (lolepezy) [reply http-client/Network/HTTP/Client/Connection.hs:208]: Ah no, I see what's wrong with it -- we create extra threads just for `tryAddress addr` which defeats the purpose.

So we can indeed do only one fork, i.e.

```
    tryAddresses result = do
        let addr : addrs = addresses
        z <- concurrently
                (tryAddress addr)
                (do
                    threadDelay connectionAttemptDelay
                    forConcurrently (zip addrs [0..]) $ \(addr, n) -> do
                        when (n > 0) $ threadDelay $ n * connectionAttemptDelay
                        tryAddress addr)
        
        let lastException = take 1 $ reverse $ filter isLeft $ uncurry (:) z
        for_ lastException $ tryPutMVar result
      where
        tryAddress addr = do
            r :: Either E.IOException a <- E.try $! cb addr
            for_ r $ \_ -> tryPutMVar result r
            pure r     
```            

Or create threads only when needed, like this.

```
    tryAddresses result = do
        lastException <- take 1 . filter isLeft <$> go addresses []
        -- If we are here then there were no successful addresses
        for_ lastException $ tryPutMVar result
      where
        go []             asyncs = mapM wait asyncs
        go (addr : addrs) asyncs =
            withAsync (tryAddress addr) $ \a -> do
                unless (null addrs) $ threadDelay connectionAttemptDelay
                go addrs $ a : asyncs

        tryAddress addr = do
            r :: Either E.IOException a <- E.try $ cb addr
            for_ r $ \_ -> tryPutMVar result r
            pure r
```            

I would prefer the latter, since it only create as many threads as actually required.

Comment (snoyberg) [reply http-client/Network/HTTP/Client/Connection.hs:208]: I'm not following this code anymore. It also seems like it's missing some pieces, like the `async` calls.

Comment (lolepezy) [reply http-client/Network/HTTP/Client/Connection.hs]: Sorry, I didn't push the code then, was just theorizing on how and when to create threads.
I pushed the commit now, with the latest version of `tryAddresses` using `withAsync`.

File: http-client/Network/HTTP/Client/Connection.hs [hunk 4]
```diff
@@ -184,10 +189,31 @@ openSocket tweakSocket addr =
...
+    -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
+    connectionAttemptDelay = 250 * 1000
+
+    tryAddresses result = do
+        lastException <- take 1 . filter isLeft <$> go addresses []
+        for_ lastException $ tryPutMVar result
+      where
+        go []             asyncs = mapM wait asyncs
+        go (addr : addrs) asyncs =
+            withAsync (E.evaluate =<< tryAddress addr) $ \a -> do
```
snoyberg, line 212: This looks like the wrong place to do an `evaluate`. Wouldn't it be better to do this inside the `try` call?

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -184,10 +189,31 @@ openSocket tweakSocket addr =
...
+                unless (null addrs) $ threadDelay connectionAttemptDelay
```
snoyberg, line 213: This seems to say that only the _final_ address won't have a delay.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -184,10 +189,31 @@ openSocket tweakSocket addr =
...
+            (\_ -> takeMVar result)
+  where
+    -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
+    connectionAttemptDelay = 250 * 1000
+
+    tryAddresses result = do
+        lastException <- take 1 . filter isLeft <$> go addresses []
+        for_ lastException $ tryPutMVar result
+      where
+        go []             asyncs = mapM wait asyncs
```
snoyberg, line 210: This logic doesn't look right to me. It seems like it's going to guarantee that all `addrs` get a forked thread.

snoyberg (changes requested):

http-client/Network/HTTP/Client/Connection.hs
lolepezy: The intention here is "if there're no more addresses to consider, don't have a delay". So if, lets say. there're 3 addresses it winds up looking like this

```
withAsync (E.evaluate =<< tryAddress addr0) $ \a0 -> do
  threadDelay connectionAttemptDelay
  withAsync (E.evaluate =<< tryAddress addr1) $ \a1 -> do
    threadDelay connectionAttemptDelay
    withAsync (E.evaluate =<< tryAddress addr2) $ \a2 ->
       -- no delay here
       mapM wait [a2, a1, a0]  
```

Unless one of `[a2, a1, a0]` fills in the `result` and then this whole stack gets cancelled by the top-level `withAsync(tryAddresses result) ... `.

http-client/Network/HTTP/Client/Connection.hs
lolepezy: Yes, pretty much. The only way to prevent it is to fill up the `result` then `go` gets interrupted. Another option would be to do something like
```
z <- isEmptyMVar result
if z 
   then go addrs $ a : asyncs
   else pure []
```
but it clutters the code and I'm not sure if it makes sense.

http-client/Network/HTTP/Client/Connection.hs
lolepezy: I added this for strictness only (to fix weird Max OS nightly tests failing), but will try to replace it with `$!` inside of the `tryAddress`.

lolepezy: And the test breaks again :|
Basically, I just need to have strict behaviour of `withAsync (do { z <- tryAddress addr; pure $! z} )` that's why `evaluate`. But I'm open to any other idea.

http-client/Network/HTTP/Client/Connection.hs
snoyberg: This is all heading back in the "too complicated" direction I originally was concerned about. I'm not convinced the code is behaving correctly, especially with the exception handling logic.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -184,10 +189,29 @@ openSocket tweakSocket addr =
...
+    -- https://datatracker.ietf.org/doc/html/rfc8305#section-5
+    connectionAttemptDelay = 250 * 1000
+
+    tryAddresses result = do
+        z <- forConcurrently (zip addresses [0..]) $ \(addr, n) -> do
+            when (n > 0) $ threadDelay $ n * connectionAttemptDelay
+            tryAddress addr
+        
+        let lastException = take 1 $ reverse $ filter isLeft z
+        for_ lastException $ tryPutMVar result
```
snoyberg, line 212: I know this is immune from a deadlock based on the first clause checking for an empty list. However, I think it would still be better to have an explicit pattern match here and generate a new exception in the case of a `Nothing`.

Also, I'm not sure that filtering for `isLeft` is an improvement. By construction, there should be no `Right` values in the list. Instead of filtering, _if_ I was going to analyze the values, I think I'd rather see an `assert` to guarantee the value is `Left`.

lolepezy: I pushed a version that does pretty much that, pattern matching and calling `error` if unexpected happens.

snoyberg (approved):
Thanks!

PR #474, opened 2022-01-11
http-client: Consume trailers and last CRLF of chunked body

### LINKED ISSUE #473: Connections are not cleaned up before reuse
I noticed this when my program was trying to start a websocket using `withConnection` right after making a call to the same server. The response before opening the websocket had returned the body using the chunked transfer encoding and the [chunked reader](https://github.com/snoyberg/http-client/blob/bc481fae6a8cf0d64ba9b4414e522e8c55c6309b/http-client/Network/HTTP/Client/Body.hs#L149-L224) doesn't read the final `CRLF` in the body. This makes the websocket client crash because it gets a `CRLF` instead of the status header. 

The problem goes away if I add `when (count == 0) $ requireNewline` [here](https://github.com/snoyberg/http-client/blob/bc481fae6a8cf0d64ba9b4414e522e8c55c6309b/http-client/Network/HTTP/Client/Body.hs#L163). But I think, in general it shouldn't be that a request is able to read (a part of) response for the previous request. So, it might be best to cleanup the connections before they go back into the pool for reuse.

Here is a little reproducer which shows this behaviour, I tested it with [this server](https://github.com/isaccanedo/http-transfer-encoding-chunked):

```haskell
module Main where

import Network.HTTP.Client
import Network.HTTP.Client.Internal
import qualified Data.ByteString.Char8 as BS

main :: IO ()
main = do
  mgr <- newManager defaultManagerSettings
  req1 mgr
  req2 mgr

req1 :: Manager -> IO ()
req1 mgr = do
  req <- parseRequest "GET http://localhost:3000"
  withResponse req mgr $ (\res -> do
                             bss <- brConsume (responseBody res)
                             BS.putStrLn $ BS.concat bss
                         )

req2 :: Manager -> IO ()
req2 mgr = do
  req <- parseRequest "GET http://localhost:3000"
  -- Read response without writing the request!
  withConnection req mgr $ (\conn -> print =<< connectionRead conn)
````

akshaymankar:
Fixes #473

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog for http-client
 
+## 0.7.10
+
+* Consume trailers and last CRLF of chunked body. The trailers are not exposed,
+  unless the raw body is requested.
+
 ## 0.7.9
 
 * Exceptions from streamed request body now cause the request to fail. Previously they were
--- a/http-client/Network/HTTP/Client/Body.hs
+++ b/http-client/Network/HTTP/Client/Body.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE MultiWayIf #-}
 module Network.HTTP.Client.Body
     ( makeChunkedReader
     , makeLengthReader
@@ -166,8 +167,11 @@ makeChunkedReader cleanup raw conn@Connection {..} = do
                 else return (empty, count0)
         if count <= 0
             then do
+                -- count == -1 indicates that all chunks have been consumed
                 writeIORef icount (-1)
-                return $ if count /= (-1) && raw then rawCount else empty
+                if | count /= -1 && raw -> S.append rawCount <$> readTrailersRaw
+                   | count /= -1        -> consumeTrailers *> pure empty
+                   | otherwise          -> pure empty
             else do
                 (bs, count') <- readChunk count
                 writeIORef icount count'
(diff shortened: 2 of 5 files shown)


Yuras:
@akshaymankar Let's consume the trailer and the final CRLF as we agreed in #473 instead of clearing the unread data

akshaymankar:
> Let's consume the trailer and the final CRLF as we agreed in #473

I've added code for this now.

> instead of clearing the unread data

There are some tests which show that there could be extra data after consuming the request. Do you think this can be safely ignored? If so I can remove the first commit.

Yuras:
Thank you, I'll take closer look tonight.

> There are some tests which show that there could be extra data after consuming the request. Do you think this can be safely ignored? If so I can remove the first commit.

I'm assuming you are referring to things like this: https://github.com/snoyberg/http-client/blob/bc481fae6a8cf0d64ba9b4414e522e8c55c6309b/http-client/test-nonet/Network/HTTP/Client/ResponseSpec.hs#L62
In this test case `HTTP/1.1` at the very end is part of the next response (modulo missing CRLF), so it should **not** be consumed. It seems to be related to HTTP pipelining: one can send HTTP request and immediately (without waiting for the response) send the next one, in which case we can receive two responses at the same time, potentially in one TCP packet.
So I'd definitely suggest to revert [the first commit](https://github.com/snoyberg/http-client/pull/474/commits/9414c434362e1d5125f3c5e4e4324782592d8d32) unless there is a good reason not to. The second commit alone seems to be enough to fix the issue reported in #473.

akshaymankar:
> In this test case `HTTP/1.1` at the very end is part of the next response (modulo missing CRLF), so it should **not** be consumed. It seems to be related to HTTP pipelining: one can send HTTP request and immediately (without waiting for the response) send the next one, in which case we can receive two responses at the same time, potentially in one TCP packet.
> So I'd definitely suggest to revert [the first commit](https://github.com/snoyberg/http-client/pull/474/commits/9414c434362e1d5125f3c5e4e4324782592d8d32) unless there is a good reason not to. The second commit alone seems to be enough to fix the issue reported in #473.

Ah, I wasn't aware of HTTP pipelining, thanks! I will delete the first commit.

Yuras (approved):
Looks good to me, thank you!
I guess @snoyberg will ask you to make a minor version bump and add a note in the changelog.

akshaymankar:
> I guess @snoyberg will ask you to make a minor version bump and add a note in the changelog.

Done.

snoyberg:
Thank you both!

PR #477, opened 2022-01-26
Allow making requests to raw IPv6 hosts (fixes #476)

### LINKED ISSUE #476: Make request to Ipv6 address
I've been trying to make a GET request to an IPv6 address but it says `invalid url`. I've also tried to use square brackets around the IP, in this case it parses the address as a `hostname` instead of `addrInfo`.

How do we make a request to an IPv6 address?

robx:
IPv6 host addresses occur in URLs in square brackets e.g. `http://[::1]`, while `Network.Socket.getAddrInfo` expects the address without brackets. This would cause requests to such URLs to fail at the name resolution stage.

This PR adds a helper to strip the brackets, and uses it in several places, notably before passing the host name to `getAddrInfo`.

Ideally, the difference between URI host name components and and host names as passed to name resolution would be made explicit in the types. It seems unlikely we want to go there, however. To gain some confidence in the change while having `String` everywhere, this PR comes in a series of commits, first introducing a newtype wrapper, then sorting the conversion into stripped and unstripped, and then removing the newtype wrapper again. This was done across the three packages `http-client`, `http-client-tls` and `http-client-openssl`. Best reviewed commit by commit.

I'm not 100% sure about the correct thing to do in each case, and have based decisions on
- In URIs, IPv6 addresses are bracketed
- In HTTP, IPv6 addresses are bracketed
- Elsewhere (network stack, SSL library) they are not bracketed
- If it's not clearly an IPv6 address, don't change behaviour

There are some unit tests for the stripping function, and a disabled test for an HTTP connection to example.com's IPv6 address.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -135,7 +135,7 @@ socketConnection socket chunksize = makeConnection
 
 openSocketConnection :: (Socket -> IO ())
                      -> Maybe HostAddress
-                     -> URIHostName String
+                     -> String -- ^ host
                      -> Int -- ^ port
                      -> IO Connection
 openSocketConnection f = openSocketConnectionSize f 8192
@@ -143,7 +143,7 @@ openSocketConnection f = openSocketConnectionSize f 8192
 openSocketConnectionSize :: (Socket -> IO ())
                          -> Int -- ^ chunk size
                          -> Maybe HostAddress
-                         -> URIHostName String
+                         -> String -- ^ host
                          -> Int -- ^ port
                          -> IO Connection
 openSocketConnectionSize tweakSocket chunksize hostAddress' host' port' =
@@ -170,7 +170,7 @@ strippedHostName hostName =
 
 withSocket :: (Socket -> IO ())
            -> Maybe HostAddress
-           -> URIHostName String
+           -> String -- ^ host
            -> Int -- ^ port
            -> (Socket -> IO a)
            -> IO a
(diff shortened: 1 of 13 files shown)


robx:
The test is failing, possibly because there's no IPv6 networking available on the CI machines?

snoyberg:
I don't have insight into how to set up the CI machines for IPv6

robx:
> I don't have insight into how to set up the CI machines for IPv6

In that case best to just leave out the test case?

snoyberg:
Unfortunately yes. Maybe marking as skipped would be good so it can be enabled in the future.

robx:
I've rewritten the PR to gain a bit more confidence that the change is correct, description updated with details. Stripping now happens in a few more places.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -11,6 +11,7 @@ module Network.HTTP.Client.Connection
     , makeConnection
     , socketConnection
     , withSocket
+    , strippedHostName
```
robx, line 14: Not really sure where this belongs in the module hierarchy.

snoyberg: This seems fine to me.

snoyberg (approved):
Awesome thanks!

robx:
@snoyberg Thanks for the quick merge and release! Could you make a new release for `http-client-tls`, too?

snoyberg:
Done

PR #480, opened 2022-01-27
Align withResponse and withResponseHistory's exception handling, and catch "resource vanished"

### LINKED ISSUE #479: Sporadic test failure: withResponseHistory and redirect
Running the test in a loop, this fails more than 10% of the time locally (macos, ghc 8.10.7).

```
Network.HTTP.Client
  Client
    withResponseHistory and redirect [✘]

Failures:

  test-nonet/Network/HTTP/ClientSpec.hs:241:18: 
  1) Network.HTTP.Client.Client withResponseHistory and redirect
       uncaught exception: IOException of type ResourceVanished
       Network.Socket.recvBuf: resource vanished (Connection reset by peer)
```

robx:
This fixes #479.

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## 0.7.11
+
+* Catch "resource vanished" exception on initial response read [#480](https://github.com/snoyberg/http-client/pull/480)
+
 ## 0.7.10
 
 * Consume trailers and last CRLF of chunked body. The trailers are not exposed,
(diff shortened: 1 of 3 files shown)


snoyberg (changes requested):
LGTM. Can you add a changelog entry describing the improvement here? Thanks!

robx:
Added a Changelog.

I'm a little bit unsure if this change isn't just papering over some deeper issues here. I can't really make sense of how connections that go away are (meant to be) dealt with. It also feels like this change breaks the `Connection` abstraction a bit, since it seems specific to `socketConnection`...

snoyberg (approved):
Thanks!

PR #484, opened 2022-02-08
Fix build on GHC 8.2

robx:
This fixes CI failing on master due to use of `<>` for string concatenation.

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -210,7 +210,7 @@ openSocket tweakSocket addr =
 
 -- Pick up an IP using an approximation of the happy-eyeballs algorithm:
 -- https://datatracker.ietf.org/doc/html/rfc8305
--- 
+--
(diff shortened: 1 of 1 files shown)


snoyberg (approved):

juhp:
Would it be possible to have a release for this, please?

Or revision for 0.7.1 (base >= 4.11).

juhp:
@robx ^^ when you have time - I hit this again ;-(

ysangkok:
@juhp What are you proposing? A revision for http-client such that it requires base 4.11 (which has `(<>)` in Prelude)? I could do that. But is it not a problem that the latest http-client-tls then becomes unbuildable on GHC 8.2? Because the latest revision of http-client-tls requires http-client 0.7.11.

ysangkok:
Ok, I have published a new revision for http-client 0.7.11 since there is no point in having the incorrect bound for that version. We'd require a new release to include this patch, and it would again be compatible with GHC 8.2.

Issue #486, opened 2022-02-28
Non-compiling example


zloilya:
In the example **Posting JSON to a server**

```
-- >   let requestObject = object
-- >    [ "name" .= ("Michael" :: Text)
-- >    , "age"  .= (30 :: Int)
-- >    ]
```

we have to add 4 extra spaces

```
-- >   let requestObject = object
-- >        [ "name" .= ("Michael" :: Text)
-- >        , "age"  .= (30 :: Int)
-- >        ]
```

and now `requestObject` below block`[`

snoyberg:
Addressed in #487, thanks!

PR #487, opened 2022-02-28
add 4 extra spaces to example on http client

zloilya:
This is the solution for [issue486](https://github.com/snoyberg/http-client/issues/486)


snoyberg (approved):
Thanks!

PR #488, opened 2022-07-18
remove original request from Response

avieth:
Deliver it instead by way of the Manager's response modifier.

See https://github.com/snoyberg/http-client/pull/464

This patch accomplishes the same thing as that one in essentially the
same way, but improves upon it by not requiring a request to exist in
order to construct a response. This can be useful in e.g. testing /
mocking definitions in which we want to express a base fixed response.
In other words, it is "less breaky" and easier to adopt.

### DIFF
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -183,7 +183,6 @@ module Network.HTTP.Client
     , responseHeaders
     , responseBody
     , responseCookieJar
-    , getOriginalRequest
     , throwErrorStatusCodes
       -- ** Response body
     , BodyReader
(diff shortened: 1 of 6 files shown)


sol:
@snoyberg any chance that this will still land.   For now, I can deal with always needing a `Request` when constructing a `Response`.

However, personally, I would still prefer if I could construct a `Response` without having to provide the original `Request`.

snoyberg:
Hey @sol, reality is that I don't have much time for maintenance of this library. If you're interested in stepping in as a comaintainer, I'd be happy to add you here and on Hackage.

sol:
> Hey @sol, reality is that I don't have much time for maintenance of this library. If you're interested in stepping in as a comaintainer, I'd be happy to add you here and on Hackage.

Hey @snoyberg, works for me.

My first priority will be fixing CI, but I will be busy for the rest of the day, so gonna look into it tomorrow.

snoyberg:
Great, thanks @sol! I just added you as a collaborator. When you're ready for Hackage access, just let me know which of the packages in this repo you'd like upload rights for and your Hackage username.

sol:
@miguelclean does this work for you as an alternative for #464.

PR #491, opened 2022-07-30
Create the ability to redact custom header values to censor sensitive…

ivo-stefchev:
Create the ability to redact custom header values to censor sensitive information.

Currently only the header field "Authorization" is redacted. That is useful, but we need to add custom fields, as different platforms use different header fields for authentication.

If this is approved - I'll add PR to "req" client, where this package is used.


ivo-stefchev:
Made a mistake with the PR, sorry

PR #492, opened 2022-07-30
Create the ability to redact custom header values to censor sensitive information

ivo-stefchev:
Create the ability to redact custom header values to censor sensitive information.

Currently only the header field "Authorization" is redacted. That is useful, but we need to add custom fields, as different platforms use different header fields for authentication.

If this is approved - I'll add PR to "req" client, where this package is used.

### DIFF
--- a/http-client/Network/HTTP/Client/Request.hs
+++ b/http-client/Network/HTTP/Client/Request.hs
@@ -49,6 +49,7 @@ import Control.Applicative as A ((<$>))
 import Control.Monad (unless, guard)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Numeric (showHex)
+import qualified Data.Set as Set
 
 import Blaze.ByteString.Builder (Builder, fromByteString, fromLazyByteString, toByteStringIO, flush)
 import Blaze.ByteString.Builder.Char8 (fromChar, fromShow)
@@ -303,6 +304,7 @@ defaultRequest = Request
         , requestManagerOverride = Nothing
         , shouldStripHeaderOnRedirect = const False
         , proxySecureMode = ProxySecureWithConnect
+        , redactHeaders = Set.singleton "Authorization"
         }
 
 -- | Parses a URL via 'parseRequest_'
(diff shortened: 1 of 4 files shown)


http-client/Network/HTTP/Client/Types.hs
```diff
@@ -620,6 +620,11 @@ data Request = Request
     -- Default: Use HTTP CONNECT.
     --
     -- @since 0.7.2
+
+    , redactHeaders :: [HeaderName]
```
snoyberg, line 624: A `Set` would probably be a better data structure for this.

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -657,9 +662,11 @@ instance Show Request where
         , "}"
         ]
 
-redactSensitiveHeader :: Header -> Header
-redactSensitiveHeader ("Authorization", _) = ("Authorization", "<REDACTED>")
-redactSensitiveHeader h = h
+redactSensitiveHeader :: [HeaderName] -> Header -> Header
+redactSensitiveHeader _ ("Authorization", _) = ("Authorization", "<REDACTED>")
```
snoyberg, line 666: I'm not sure I agree with having `Authorization` sit separately from the rest of these. I'd lean towards having the default `redactHeaders` be a singleton set of just `"Authorization"`

http-client/http-client.cabal
```diff
@@ -1,5 +1,5 @@
 name:                http-client
-version:             0.7.12
+version:             0.7.12.1
```
snoyberg, line 2: This would be a minor version bump to 0.7.13, and require a ChangeLog entry too.

http-client/Network/HTTP/Client/Request.hs
```diff
@@ -303,6 +304,7 @@ defaultRequest = Request
         , requestManagerOverride = Nothing
         , shouldStripHeaderOnRedirect = const False
         , proxySecureMode = ProxySecureWithConnect
+        , redactHeaders = Set.fromList [ "Authorization" ]
```
snoyberg, line 307: Really pedantic comment, but `Set.singleton "Authorization"` would be more idiomatic and likely a tiny bit more efficient.

http-client/http-client.cabal
```diff
@@ -1,5 +1,5 @@
 name:                http-client
-version:             0.7.12
+version:             0.7.13
```
snoyberg, line 2: Please also include a ChangeLog entry for this.

snoyberg (changes requested):

snoyberg:
LGTM, thanks!

PR #499, opened 2023-01-15
Haddock fixes

jhrcek:


http-client/Network/HTTP/Client/Request.hs
```diff
@@ -122,9 +122,9 @@ throwErrorStatusCodes req res = do
 -- You can place the request method at the beginning of the URL separated by a
 -- space, e.g.:
 --
--- @@@
```
jhrcek, line 125: There's no `@@@` syntax afaik. This leads to example rendering like this

Screenshot from 2023-01-15 10-44-28


After this fix:

Screenshot from 2023-01-15 10-43-15

snoyberg (approved):
Thanks!

PR #505, opened 2023-06-06
using crypton

kazu-yamamoto:
Background is written in https://github.com/yesodweb/wai/pull/931.

I would like to have new releases of `http-client` and `http-client-tls`.


kazu-yamamoto:
New `warp` and `warp-tls` are available on Hackage.

kazu-yamamoto:
@snoyberg Thank you for your effort!

PR #506, opened 2023-06-06
Move to crypton, update CI

snoyberg:
CC @kazu-yamamoto


http-conduit/http-conduit.cabal
```diff
@@ -63,7 +63,6 @@ test-suite test
                  , HUnit
                  , hspec >= 1.3
                  , data-default-class
-                 , connection >= 0.2
```
kazu-yamamoto, line 66: @snoyberg Sorry but this should be `crypton-connection`.

kazu-yamamoto:
@snoyberg Could you take time to put this forward?

snoyberg:
I'll try to take care of this, but it will probably be a few days at least.

PR #510, opened 2023-06-20
Add attoparsec-aeson dependency

9999years:
This will enable compatibility with aeson-2.2 when it's released. aeson-2.2 moves the `Data.Aeson.Parser` module into a separate package (attoparsec-aeson).

### DIFF
--- a/http-conduit/http-conduit.cabal
+++ b/http-conduit/http-conduit.cabal
@@ -42,6 +42,7 @@ library
 
     if flag(aeson)
       build-depends: aeson                 >= 0.8
+                   , attoparsec-aeson      >= 2.2
 
     if !impl(ghc>=7.9)
       build-depends:   void >= 0.5.5


snoyberg:
This package doesn't exist on Hackage. Is there a reason for these PRs at this point?

9999years:
We wanted to use the unreleased versions internally for the new [optional field support](https://github.com/haskell/aeson/pull/1039), and referencing PRs in our build is a bit nicer than committing a bunch of patch files. We don't need to merge this until aeson-2.2 is released on Hackage (but this PR will be ready when it is).

axelclark:
I'm getting this error when installing `http-conduit` today.
```haskell
Installing   aeson-2.2.0.0 (lib)
Completed    aeson-2.2.0.0 (lib)
...
Network/HTTP/Simple.hs:106:1: error:
    Could not find module ‘Data.Aeson.Parser’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
    |
106 | import Data.Aeson.Parser (json')
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build http-conduit-2.3.8.2. See the build log above for
details.
```

snoyberg:
I can't merge this pull request as-is. The dependency on attoparsec-aeson will force using only newer versions of aeson, causing unnecessary complications for upgrading existing codebases. This will require some fancy work with flags to make the dependency contingent on the version of `aeson` being used.

endgame:
https://github.com/haskell/aeson/issues/1048 should fix this without fancy flag work.

endgame:
https://github.com/haskell/aeson/issues/1048 has been completed, and the compatibility package released.

juhp:
Maybe someone can re-run the CI? :)

snoyberg:
Rerunning won't fix it, the `stack.yaml` file needs to be updated to include the new dependency.

http-conduit/http-conduit.cabal
```diff
@@ -42,6 +42,7 @@ library
 
     if flag(aeson)
       build-depends: aeson                 >= 0.8
+                   , attoparsec-aeson      >= 2.2
```
ysangkok, line 45: This needs to be lowered to allow version 2.1 which is the empty package compatible with older aesons.

endgame:
It would be great if this could receive prompt attention — amazonka-2.0 won't release until we are able to lift our `aeson` upper bound, and we can't do that because of the `http-conduit` breakage.

PR #515, opened 2023-08-17
catching up to tls 1.8.0

kazu-yamamoto:
`tls` v1.8.0 (which will be eventually released) does not throw `TLSError`.
It is not an instance of `Exception` anymore.
This PR supports both old and new versions of `tls`.

For more information, please read: https://github.com/haskell-tls/hs-tls/pull/457

Cc: @andrewthad


snoyberg (approved):

snoyberg:
Thanks! Released to Hackage as 0.3.6.3.

PR #523, opened 2023-12-19
Handle HTTP 103 Early Hints

### LINKED ISSUE #521: Misbehavior in the presence of HTTP 103 early hints
Initially I filed https://github.com/fpco/http-reverse-proxy/issues/44, but then I decided to test if `http-client` itself works properly in the presence of HTTP 103 early hints, and it seems it does not.

I added another repro alongside my existing one, so you can check it as follows. The meat of it is [here](https://github.com/codedownio/http-reverse-proxy/blob/http-103-early-hints-repro/app-client/Main.hs).

```
cd $(mktemp -d)
git clone git@github.com:codedownio/http-reverse-proxy.git --branch http-103-early-hints-repro
npm install && npm run start # Starts a test Node.js server

# Now switch to the same folder in another terminal and do
stack run client-test
```

This will use `http-client` to make a normal request, as well as an early-hints request.

## Normal request output

```
Doing an HTTP request to http://localhost:3005/simple
The status code was: 200
Full response: Response {responseStatus = Status {statusCode = 200, statusMessage = "OK"}, responseVersion = HTTP/1.1, responseHeaders = [("X-Powered-By","Express"),("Content-Type","text/html; charset=utf-8"),("Content-Length","16"),("ETag","W/\"10-uViC3pa4b5/U2h5JkCXIHUyMvk0\""),("Date","Wed, 22 Nov 2023 07:25:23 GMT"),("Connection","keep-alive"),("Keep-Alive","timeout=5")], responseBody = "Simple response!", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
"Simple response!"
```

## Early hints request output

```
Doing an HTTP request to http://localhost:3005/early_hints
The status code was: 103
Full response: Response {responseStatus = Status {statusCode = 103, statusMessage = "Early Hints"}, responseVersion = HTTP/1.1, responseHeaders = [("Link","</some-file.js>")], responseBody = "", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
""
```

It seems that `http-client` sees the `HTTP 103` status line and thinks it's at the end of the request. It gets the headers, but doesn't proceed to get the final status code or the response body.

thomasjm:
This fixes #521 by handling the 103 Early Hints header specially.

It doesn't quite fix https://github.com/fpco/http-reverse-proxy/issues/44. To make the reverse proxy work properly in the present of 103 Early Hints, I think we need to separate the early hint headers from the normal headers. I don't think that can be done without changing the API. I'm thinking we can add a `responseEarlyHeaders` field to the `Response` object and put them in there instead?

### DIFF
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -5,6 +5,7 @@ module Network.HTTP.Client.Connection
     ( connectionReadLine
     , connectionReadLineWith
     , connectionDropTillBlankLine
+    , connectionUnreadLine
     , dummyConnection
     , openSocketConnection
     , openSocketConnectionSize
@@ -59,6 +60,11 @@ connectionReadLineWith mhl conn bs0 =
             (x, S.drop 1 -> y) -> do
                 unless (S.null y) $! connectionUnread conn y
                 return $! killCR $! S.concat $! front [x]
+
+connectionUnreadLine :: Connection -> ByteString -> IO ()
+connectionUnreadLine conn line = do
+  connectionUnread conn (S.pack [charCR, charLF])
+  connectionUnread conn line
 
 charLF, charCR :: Word8
 charLF = 10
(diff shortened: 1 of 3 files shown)


snoyberg (approved):

snoyberg:
I think such an addition can make sense. I think it can also be done in a mostly backwards-compatible way.

For this PR, can you add a version bump and changelog entry?

thomasjm:
Closing in favor of #524 per discussion there.

PR #524, opened 2023-12-19
Move early hints into `responseEarlyHints` field

thomasjm:
This is a continuation of #523, developing the idea that we can put the early hints into a new field of the `Response` object called `responseEarlyHints`.

AFAICT this is not a breaking change, since the `Response` data constructors are not exposed and the `StatusHeaders` type is only exposed in an internal module.

### DIFF
--- a/http-client/Network/HTTP/Client/Types.hs
+++ b/http-client/Network/HTTP/Client/Types.hs
@@ -635,7 +635,7 @@ data Request = Request
     --
     -- @since 0.7.13
 
-    , earlyHintHeaderReceived :: Header -> IO ()
+    , earlyHintHeadersReceived :: [Header] -> IO ()
     -- ^ Called every time an HTTP 103 Early Hints header is received from the server.
     --
     -- @since 0.7.16
--- a/http-client/Network/HTTP/Client/Types.hs
+++ b/http-client/Network/HTTP/Client/Types.hs
@@ -90,7 +90,7 @@ data Connection = Connection
     }
     deriving T.Typeable
 
-data StatusHeaders = StatusHeaders Status HttpVersion RequestHeaders
+data StatusHeaders = StatusHeaders Status HttpVersion RequestHeaders RequestHeaders
     deriving (Show, Eq, Ord, T.Typeable)
 
 -- | A newtype wrapper which is not exported from this library but is an
@@ -715,6 +715,11 @@ data Response body = Response
     -- via @getOriginalRequest@ instead.
     --
     -- Since 0.7.8
+    , responseEarlyHints :: ResponseHeaders
+    -- ^ Early response headers sent by the server, as part of an HTTP
+    -- 103 Early Hints section.
+    --
+    -- Since 0.7.16
     }
     deriving (Show, T.Typeable, Functor, Data.Foldable.Foldable, Data.Traversable.Traversable)
 
--- a/http-client/Network/HTTP/Client/Types.hs
+++ b/http-client/Network/HTTP/Client/Types.hs
@@ -636,7 +636,7 @@ data Request = Request
     -- @since 0.7.13
 
     , earlyHintHeadersReceived :: [Header] -> IO ()
-    -- ^ Called every time an HTTP 103 Early Hints header is received from the server.
+    -- ^ Called every time an HTTP 103 Early Hints header section is received from the server.
     --
     -- @since 0.7.16
     }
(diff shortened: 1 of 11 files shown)


thomasjm:
Hmm, I just realized this is not going to actually provide a performance improvement, because `parseStatusHeaders` won't return until it actually reads the final status + headers, at which time the response is presumably ready to go anyway.

This limits the usefulness of this when it comes to implementing performant reverse proxying, but at least it fixes the behavior on #521.

It seems like it will require more of an API change in order to first return early hints headers, and then at a later time return the actual status + headers + response.

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -715,6 +715,11 @@ data Response body = Response
     -- via @getOriginalRequest@ instead.
     --
     -- Since 0.7.8
+    , responseEarlyHints :: ResponseHeaders
+    -- ^ Early response headers sent by the server, as part of an HTTP
+    -- 103 Early Hints section.
+    --
+    -- Since 0.7.16
```
snoyberg, line 722: In the other PR I asked for a version bump, but if you can make this the new version and add a changelog/cabal file change, that would be perfect.

thomasjm: :+1:  This PR slightly improves on some aspects of the other one. I'll just close the other one.

thomasjm:
I just added the cabal version bump and changelog update.

I'm still not sure how the early hint headers can be exposed in a way that the user can access them immediately after they arrive, as when you're trying to e.g. write a reverse proxy. When you look at the key functions of `http-client` such as `responseOpen`:

```haskell
responseOpen :: Request -> Manager -> IO (Response BodyReader)
```

it seems like it would have to change to something like this:

```haskell
-- An extra callback that receives any early hint headers
responseOpen' :: Request -> Manager -> ([Header] -> IO ()) -> IO (Response BodyReader)
```

OR, less invasively, how about being able to specify a `[Header] -> IO ()` callback on the `Request` object? It already has a couple IO callbacks.

http-client/Network/HTTP/Client/Types.hs
thomasjm: Done

snoyberg:
I think adding a callback to the `Request` object makes perfect sense in a case like this. Do you want to include that in this PR before I merge?

thomasjm:
I might as well include it, it looks like I'll have to plumb the callback through most of this.

thomasjm:
Okay, I've added the callback.

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -634,6 +634,11 @@ data Request = Request
     -- ^ List of header values being redacted in case we show Request.
     --
     -- @since 0.7.13
+
+    , earlyHintHeaderReceived :: Header -> IO ()
```
snoyberg, line 638: This signature surprises me a bit, I'd have thought we'd send all the headers at once. There shouldn't be a significant time delay in slurping in all the headers, and from the user's perspective there's no way to distinguish if all early hint headers have been received or not given the single-header API.

thomasjm: I was thinking it's theoretically possible for a web server to be doing some streaming server-side rendering, and outputting "Link" headers one by one as it encounters them. So then there might be time delays between the different headers.

However, it seems like the current [Node.js](https://nodejs.org/api/http.html#responsewriteearlyhintshints-callback) implementation sends them as a batch.

You're right, the user does need to know when they've all been received. I'd suggest one of the following:
a) `Maybe Header -> IO ()` (sends a `Just` on every header, then `Nothing` to indicate completion)
b) `[Header] -> IO ()` (sends them all as a batch)

Option b) was my original plan and would probably be fine, and would make it easier to write a reverse proxy. It would be messy to change our minds later though.

The [RFC](https://www.rfc-editor.org/rfc/rfc8297#section-2) says the following, which suggests to me that a server which wishes to do streaming rendering can just send multiple early hint batches.

```
A server MAY emit multiple 103 (Early Hints) responses with
additional header fields as new information becomes available while
the request is being processed.
```

snoyberg: I think (b) is fine, I'm not aware of any other API where individual headers can be streamed, so most likely the multiple 103s approach would be used if that needs to work reliably.

thomasjm:
Okay, the callback signature is `[Header] -> IO ()` and I added a test of multiple 103 header sections.

snoyberg (approved):

snoyberg:
Merged and released, thanks!

PR #527, opened 2024-02-07
Add low level createConnection/disposeConnection as an alternative to withConnection

thomasjm:
I basically needed these functions for UnliftIO reasons. I wanted to use `http-client` to access a WebSocket endpoint, where I need to read and write interactively, as you would with [withConnection](https://hackage.haskell.org/package/http-client-0.7.16/docs/src/Network.HTTP.Client.Core.html#withConnection).

However, `withConnection` wasn't cutting it, because it runs in the `IO` monad, while I want to use a more general monad. IIRC I wasn't able to use MonadUnliftIO to get around it. So I think it would be nice to expose these functions, so users can do general bracket-style stuff with them.

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## 0.7.17
+
+* Expose low-level createConnection/disposeConnection functions.
+
 ## 0.7.16
 
 * Add `responseEarlyHints` field to `Response`, containing a list of all HTTP 103 Early Hints headers received from the server.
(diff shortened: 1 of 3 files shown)


snoyberg:
Does this do what you're looking for?

```haskell
liftedWithConnection :: MonadUnliftIO m => Request -> Manager -> (Connection -> m a) -> m a
liftedWithConnection req manager inner =
    withRunInIO $ \run -> withConnection req manager (run . inner)
```

thomasjm:
Actually yes, I was able to make my monad be `MonadUnliftIO` and this works. Thanks!

IIRC an earlier version of my code used a monad that was incompatible with `MonadUnliftIO`, so maybe there's still a use-case for separate functions. But I'm happy now--sorry for the noise.

PR #529, opened 2024-03-07
log requestBody in Request

stevemao:
This is very useful for debugging

### DIFF
--- a/http-client/Network/HTTP/Client/Types.hs
+++ b/http-client/Network/HTTP/Client/Types.hs
@@ -665,7 +665,7 @@ instance Show Request where
         , "  requestHeaders       = " ++ show (DL.map (redactSensitiveHeader $ redactHeaders x) (requestHeaders x))
         , "  path                 = " ++ show (path x)
         , "  queryString          = " ++ show (queryString x)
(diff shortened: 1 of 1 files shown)


http-client/Network/HTTP/Client/Types.hs
```diff
@@ -665,7 +665,7 @@ instance Show Request where
         , "  requestHeaders       = " ++ show (DL.map (redactSensitiveHeader $ redactHeaders x) (requestHeaders x))
         , "  path                 = " ++ show (path x)
         , "  queryString          = " ++ show (queryString x)
-        --, "  requestBody          = " ++ show (requestBody x)
+        , "  requestBody          = " ++ show (requestBody x)
```
sol, line 668: `RequestBody` does not have a `Show` instance, so I don't think this is possible, at least not in the general case.

sol:
What you technically could do is pattern match on the request body and show it only when it is `RequestBodyLBS`, `RequestBodyBS`, and possibly `RequestBodyBuilder`.  Not sure if that's the right thing to do.

Anyway, the PR in its current form I think does not compile.  I'm gonna close this.

stevemao:
Good point. Can we remove this comment then 😄

PR #539, opened 2024-07-05
http-conduit: Store the cleanup actions of responses in `ResourceT`

### LINKED ISSUE #537: http-conduit: Space leaks when using wide `ResourceT`
A colleague (@kokobd) did a lot of digging into space leaks while using `amazonka`, and we think we've chased it down to a problem in `http-conduit`.

**Problem:** `Response`s cannot be garbage collected until execution leaves the surrounding `ResourceT`, even if the response is closed. It's quite a common idiom to have very large `runResourceT` blocks, which leads to space leaks until the block is left.

The problem appears to be in `http-conduit:Network.HTTP.Conduit.http`:

https://github.com/snoyberg/http-client/blob/6f742e857a91b37f9d1618a30510fc4f97989342/http-conduit/Network/HTTP/Conduit.hs#L315-L324

When a response body is completely read, it calls `release key` which closes the connection and removes it from the `ResourceT`. However, calling `http-conduit:Network.HTTP.Conduit.responseClose` cleans up the connection itself but does not tell the enclosing `ResourceT` that the connection has been closed. `ResourceT` then holds onto the dead `Response` until the block is left, and it does its cleanup. This is now a no-op because cleanup has already happened, so we've been holding onto the `Response` for no benefit.

An example of this in the wild: [`amazonka` closes connections in this way](https://github.com/brendanhay/amazonka/blob/4873cc451113147d071721c97704ac648d71e9ee/lib/amazonka-core/src/Amazonka/Response.hs#L55) to try and release resources as soon as it can; see https://github.com/brendanhay/amazonka/issues/490 for full details.

**Possible Solution:** Instead of using `allocate` with `responseClose`, have `Response` do a `ResourceT`-based cleanup and put the real cleanup action only in the `ResourceT`. I think this works (if exceptions are properly masked):

1. Call `Client.responseOpen` to generate the response.
2. Extract the `ResponseClose` from the returned response and `register` it into the `ResourceT`, returning `key`.
3. Replace the `responseClose'` field of the response with a new `ResponseClose $ release key`.
4. (`http-conduit` already does this, listing for completeness) Attach `release key` to the `responseBody` field of the response.

Then exiting `ResourceT` still cleans up the response, but both an explicit `responseClose` and consuming the whole body will release the response more promptly.

JackKelly-Bellroy:
Because `ResourceT` needs to hold onto the `Response` so it can perform cleanup when leaving `runResourceT`, move the cleanup action from the `Response` itself and register it with `ResourceT`. Then all code paths which trigger cleanup (leaving `ResourceT`, consuming the body, calling `responseClose`) will do so by releasing the `Response` from the `ReleaseMap`, preventing a space leak.

Fixes #537

### DIFF
--- a/http-conduit/ChangeLog.md
+++ b/http-conduit/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for http-conduit
 
+## Unreleased
+
+* Fix space leaks when closing responses [#539](https://github.com/snoyberg/http-client/pull/539)
+
 ## 2.3.8.3
 
 * aeson 2.2 support [#512](https://github.com/snoyberg/http-client/pull/512)
--- a/http-conduit/Network/HTTP/Conduit.hs
+++ b/http-conduit/Network/HTTP/Conduit.hs
@@ -233,7 +233,7 @@ import           Control.Applicative          as A ((<$>))
 import           Control.Monad.IO.Unlift      (MonadIO (liftIO))
 import           Control.Monad.Trans.Resource
 
-import qualified Network.HTTP.Client          as Client (httpLbs, responseOpen, responseClose)
+import qualified Network.HTTP.Client          as Client (httpLbs, responseOpen)
 import qualified Network.HTTP.Client          as HC
 import qualified Network.HTTP.Client.Conduit  as HCC
 import           Network.HTTP.Client.Internal (createCookieJar,
@@ -244,9 +244,8 @@ import           Network.HTTP.Client.Internal (Manager, ManagerSettings,
                                                managerTlsConnection, newManager)
 import           Network.HTTP.Client          (parseUrl, parseUrlThrow, urlEncodedBody, applyBasicAuth,
                                                defaultRequest, parseRequest, parseRequest_)
-import           Network.HTTP.Client.Internal (addProxy, alwaysDecompress,
-                                               browserDecompress)
-import           Network.HTTP.Client.Internal (getRedirectedRequest)
+import           Network.HTTP.Client.Internal (ResponseClose (..), addProxy, alwaysDecompress,
+                                               browserDecompress, getRedirectedRequest)
 import           Network.HTTP.Client.TLS      (mkManagerSettings,
                                                tlsManagerSettings)
 import           Network.HTTP.Client.Internal (Cookie (..), CookieJar (..),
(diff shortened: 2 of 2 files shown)


JackKelly-Bellroy:
We're testing this for now, but will mark it as ready once we're sure it works.

kokobd:
Here is the test program.

```haskell
{-# LANGUAGE OverloadedLabels #-}

module Main where

import Amazonka qualified
import Amazonka.S3 qualified as S3
import Control.Monad (void)
import Control.Monad.Random.Class
import Control.Monad.Trans.Class (lift)
import Data.Generics.Labels ()
import Data.Text qualified as T
import Data.Traversable (for)

main :: IO ()
main = do
  env <- Amazonka.newEnv Amazonka.discover
  void . Amazonka.runResourceT $
    for (replicate 1000 ()) $ \_ -> do
      randomStr <- fmap (T.pack . take 10000000) . lift $ getRandomRs ('a', 'z')
      lift $ putStrLn "Sending 10MB of data"
      void . Amazonka.send env $
        S3.newPutObject "bellroy-eventlog-test" "test" (Amazonka.toBody randomStr)
```

Test command: `AWS_PROFILE=xxx AWS_REGION=us-east-2 cabal run --enable-profiling exe:amazonka-profiling -- +RTS -l -hT`

Below are eventlog htmls before and after this PR:
image
image

JackKelly-Bellroy:
This definitely improves things, so I'm un-marking it as draft.

snoyberg:
I'm not following the change here. The existing `allocate` call should be doing exactly the same registration with `ResourceT`. Is it possible that this is a subtle laziness bug that the rewrite is fixing? I'm hesitant to merge this as-is given that I don't understand why it helps.

JackKelly-Bellroy:
In `http-conduit` as it currently stands, calling `responsClose` on a `Response` will close the connection, but you leak space because `ResourceT` needs to hold onto it until execution leaves the enclosing `runResourseT`. This comes up in e.g., a bunch of `amazonka` code, where it optimistically closes connections that don't return streaming responses, but you get space leaks until execution leaves the `ResourceT`.

This PR changes that by replacing the `responseClose'` field on the connection, so if someone calls `responseClose`, it asks `ResourceT` to release the connection. This means closed `Response`s can be GC'd before execution has left the enclosing `ResourceT`.

We can't use `allocate (Client.responseOpen req man) Client.responseClose` because we don't want `Client.responseClose` to be the registered cleanup action — that would cause an infinite loop. We want to register the cleanup action that's initially in the `Response` with `ResourceT`. Hence:

1. `Client.responseOpen` to create the `Response`
2. Extract the cleanup action from the `Response` and register it with `ResourceT`.
3. Replace the cleanup action in the `Response` with one that `releases` the registered action.
4. (As in current `master`) Release the registered action if the response conduit is completely consumed.

snoyberg:
Ah, got it, thanks for the explanation. Makes sense, and this looks good! Merging.

PR #540, opened 2024-07-05
Force response body update

### LINKED ISSUE #538: http-conduit: `Request`s may be retained in `Response`s
A colleague (@kokobd) did a lot of digging into space leaks while using amazonka, and we've observed that large requests cause a noticeable space leak. This sounds similar to https://github.com/brendanhay/amazonka/issues/475

When `http-conduit` constructs a response, it stores the original request in a record field and overwrites its body:

https://github.com/snoyberg/http-client/blob/6f742e857a91b37f9d1618a30510fc4f97989342/http-client/Network/HTTP/Client/Response.hs#L164

We suspect that this record update is never forced (there are no bang-patterns or other strictness tools that we can see), so the original request body cannot be garbage collected until execution leaves the `ResourceT`.

It might be necessary to add bang patterns to (at least) the `responseOriginalRequest` field of `Response` and the `requestBody` field of `Request`. It is probably safe for the fields of the `RequestBody` constructors to remain lazy.

JackKelly-Bellroy:
Fixes #538 

Draft for now, but keen for comments: do you prefer strict `let`s? `seq`? Bang patterns?

We will test this internally before recommending that it gets merged.

### DIFF
--- a/http-client/ChangeLog.md
+++ b/http-client/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for http-client
 
+## Unreleased
+
+* Fix a space leak when storing original requests inside responses [#540](https://github.com/snoyberg/http-client/pull/540)
+
 ## 0.7.17
 
 * Add `managerSetMaxHeaderLength` to `Client` to change `ManagerSettings` `MaxHeaderLength`.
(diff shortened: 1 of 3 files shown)


snoyberg:
Either approach is fine with me.

JackKelly-Bellroy:
We could never prove an improvement in space usage with this PR, so I'm going to close it.

PR #514, opened 2023-08-15
Allow customizing max header length

tfausak:
Fixes #502. 

I'm not sure if this is the right way to add this customization. I turned the constant into a parameter and then just kept adding parameters until the compiler was happy. Then I pushed the configuration into the `Manager` since that felt like the right place to put it. 

Some of the function signatures could be cleaner if this configuration was contained in the `Connection`, but that didn't feel like the right place for this — for whatever reason. Also I have no idea how important it is to have a stable API for these functions. As a user of `http-client`, I basically only ever use `newManager` and `httpLbs`. 

I'm very much open to feedback here! Please let me know if I can change anything to make this easier to merge.

PR #549, opened 2024-12-12
Add the possibility to set the maximum number of header fields

marinelli:
This PR adds the possibility to configure the maximum number of HTTP header fields in a message.
This is similar to what it was done for configuring the maximum size of an HTTP header field (https://github.com/snoyberg/http-client/pull/514).

### DIFF
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -113,6 +113,7 @@ module Network.HTTP.Client
     , managerSetInsecureProxy
     , managerSetSecureProxy
     , managerSetMaxHeaderLength
+    , managerSetMaxNumberHeaders
     , ProxyOverride
     , proxyFromRequest
     , noProxy
@@ -325,6 +326,11 @@ managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxHeaderLength l manager = manager
     { managerMaxHeaderLength = Just $ MaxHeaderLength l }
+
+-- @since 0.7.18
+managerSetMaxNumberHeaders :: Int -> ManagerSettings -> ManagerSettings
+managerSetMaxNumberHeaders l manager = manager
+    { managerMaxNumberHeaders = Just $ MaxNumberHeaders l }
 
 -- $example1
 -- = Example Usage
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -323,14 +323,14 @@ managerSetProxy :: ProxyOverride -> ManagerSettings -> ManagerSettings
 managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 
 -- @since 0.7.17
-managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
+managerSetMaxHeaderLength :: Word -> ManagerSettings -> ManagerSettings
 managerSetMaxHeaderLength l manager = manager
-    { managerMaxHeaderLength = Just $ MaxHeaderLength l }
+    { managerMaxHeaderLength = MaxHeaderLength l }
 
 -- @since 0.7.18
-managerSetMaxNumberHeaders :: Int -> ManagerSettings -> ManagerSettings
-managerSetMaxNumberHeaders l manager = manager
-    { managerMaxNumberHeaders = Just $ MaxNumberHeaders l }
+managerSetMaxNumberHeaders :: Word -> ManagerSettings -> ManagerSettings
+managerSetMaxNumberHeaders n manager = manager
+    { managerMaxNumberHeaders = MaxNumberHeaders n }
 
 -- $example1
 -- = Example Usage
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -323,12 +323,12 @@ managerSetProxy :: ProxyOverride -> ManagerSettings -> ManagerSettings
 managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 
 -- @since 0.7.17
-managerSetMaxHeaderLength :: Word -> ManagerSettings -> ManagerSettings
+managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxHeaderLength l manager = manager
     { managerMaxHeaderLength = MaxHeaderLength l }
 
 -- @since 0.7.18
-managerSetMaxNumberHeaders :: Word -> ManagerSettings -> ManagerSettings
+managerSetMaxNumberHeaders :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxNumberHeaders n manager = manager
     { managerMaxNumberHeaders = MaxNumberHeaders n }
 
--- a/http-client/Network/HTTP/Client.hs
+++ b/http-client/Network/HTTP/Client.hs
@@ -325,12 +325,12 @@ managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 -- @since 0.7.17
 managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxHeaderLength l manager = manager
-    { managerMaxHeaderLength = MaxHeaderLength l }
+    { managerMaxHeaderLength = Just $ MaxHeaderLength l }
 
 -- @since 0.7.18
 managerSetMaxNumberHeaders :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxNumberHeaders n manager = manager
-    { managerMaxNumberHeaders = MaxNumberHeaders n }
+    { managerMaxNumberHeaders = Just $ MaxNumberHeaders n }
 
 -- $example1
 -- = Example Usage
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -50,7 +50,7 @@ connectionReadLineWith mhl conn bs0 =
     go bs front total =
         case S.break (== charLF) bs of
             (_, "") -> do
-                let total' = total + fromIntegral (S.length bs)
+                let total' = total + S.length bs
                 when (total > unMaxHeaderLength mhl) $ do
                     -- We reached the maximum length for an header field.
                     throwHttp OverlongHeaders
--- a/http-client/Network/HTTP/Client/Connection.hs
+++ b/http-client/Network/HTTP/Client/Connection.hs
@@ -31,32 +31,31 @@ import Data.Function (fix)
 import Data.Maybe (listToMaybe)
 import Data.Word (Word8)
 
-connectionReadLine :: Maybe MaxHeaderLength -> Connection -> IO ByteString
+connectionReadLine :: MaxHeaderLength -> Connection -> IO ByteString
 connectionReadLine mhl conn = do
     bs <- connectionRead conn
     when (S.null bs) $ throwHttp IncompleteHeaders
     connectionReadLineWith mhl conn bs
 
 -- | Keep dropping input until a blank line is found.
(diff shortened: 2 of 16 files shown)


snoyberg (changes requested):
Thanks! Two pieces of feedback:

1. Can you update the version number in the cabal file too?
2. It's slightly surprising that the new config value is a `Maybe Int` but it comes with a default of 100 and doesn't appear to have a way to set it to `Nothing`. Not a strong comment, but something more consistent might make sense.

marinelli:
> Thanks! Two pieces of feedback:

Thanks for having a look :)

> 1. Can you update the version number in the cabal file too?

done

> 2. It's slightly surprising that the new config value is a `Maybe Int` but it comes with a default of 100 and doesn't appear to have a way to set it to `Nothing`. Not a strong comment, but something more consistent might make sense.

I used the same strategy of `managerMaxHeaderLength`, and for what I understood is the following:
- if you want to use set a maximum value, use something like `managerMaxNumberHeaders 123 defaultManagerSettings`
- if you want to set it to Nothing, import `Network.HTTP.Client.Internal`, and do something like: `defaultManagerSettings { managerMaxNumberHeaders = Nothing }`.

I think we introduced some extra complexity, not really required.
The `MaxHeaderLength` and the new `MaxNumberHeaders` are used in predicates where we compare those max values with a counter, we can already use a negative number to obtain the same behavior of configuring those parameters to `Nothing`, ie: `managerMaxNumberHeaders (- 1) defaultManagerSettings`.

The main purpose of configuring those parameters to `Nothing` would be for _testing_ and _debugging_ (I think), so we could just remove the `Maybe`, and have something like this: https://github.com/marinelli/http-client/compare/set-max-headers...marinelli:http-client:use-ints.

If we would like to continue to use a `Maybe` value for those max settings in `ManagerSettings`, well, I would change it to a `Maybe Word`.

cc: @tfausak

snoyberg:
I don't remember the reasons behind the max header length decisions. I'm fine with any approach. Even if we went over to `Word`, using `0` to mean "unlimited" would be reasonable too (since disallowing all headers wouldn't make sense).

@tfausak any opinion from you on which way to go?

tfausak:
I like the idea of removing `Maybe`. I also think that switching to `Word` makes sense, but that may require some annoying refactoring. 

There may be no need for a special "unlimited" value. You could just set it to `maxBound` which is effectively unlimited. But also treating `0` as "unlimited" makes sense to me. You can't really do HTTP without any headers, so setting the limit to `0` be useless.

marinelli:
> I like the idea of removing `Maybe`. I also think that switching to `Word` makes sense, but that may require some annoying refactoring.
> 
> There may be no need for a special "unlimited" value. You could just set it to `maxBound` which is effectively unlimited. But also treating `0` as "unlimited" makes sense to me. You can't really do HTTP without any headers, so setting the limit to `0` would be useless.

I had a look here https://www.rfc-editor.org/rfc/rfc2616#page-31, and it seems that header fields can also be omitted, so `0` might be a proper value. Yeah, we could use `maxBound`, but I think also negative numbers are a valid option.
@tfausak I already refactored the code here https://github.com/marinelli/http-client/compare/set-max-headers...marinelli:http-client:use-ints, and I do not think that switching to `Word`s will be so complicated.

tfausak:
Technically the RFC allows zero headers, but practically the `Host` header is always present for HTTP/1.1 requests. There's also this bit: https://www.rfc-editor.org/rfc/rfc2616#section-9

> The Host request-header field (section 14.23) MUST accompany all HTTP/1.1 requests.

marinelli:
> Technically the RFC allows zero headers, but practically the `Host` header is always present for HTTP/1.1 requests. There's also this bit: https://www.rfc-editor.org/rfc/rfc2616#section-9
> 
> > The Host request-header field (section 14.23) MUST accompany all HTTP/1.1 requests.

I don't know, it's a bit confusing to me to assign to `0` the meaning of _unlimited_ value, I could think to use `0` to test server implementations.
Anyway, if we will find that using `0` with that meaning might be confusing, we can just change it in an other PR.

marinelli:
> > Technically the RFC allows zero headers, but practically the `Host` header is always present for HTTP/1.1 requests. There's also this bit: https://www.rfc-editor.org/rfc/rfc2616#section-9
> > > The Host request-header field (section 14.23) MUST accompany all HTTP/1.1 requests.
> 
> I don't know, it's a bit confusing to me to assign to `0` the meaning of _unlimited_ value, I could think to use `0` to test server implementations. Anyway, if we will find that using `0` with that meaning might be confusing, we can just change it in an other PR.

Here it is https://github.com/marinelli/http-client/pull/1.

http-client/Network/HTTP/Client/Connection.hs
```diff
@@ -31,32 +31,31 @@ import Data.Function (fix)
...
     go bs0 id 0
   where
     go bs front total =
         case S.break (== charLF) bs of
             (_, "") -> do
-                let total' = total + S.length bs
-                case fmap unMaxHeaderLength mhl of
-                    Nothing -> pure ()
-                    Just n -> when (total' > n) $ throwHttp OverlongHeaders
+                when (total >= unMaxHeaderLength mhl && total /= 0) $ do
```
tfausak, line 53: This needs to compare against the new total, not the old one. And it should be checking to see if the `MaxHeaderLength` is 0, not the total. 
```suggestion
                let n = unMaxHeaderLength mhl
                    total' = total + fromIntegral (S.length bs)
                when (total' >= n && n /= 0) $ do
```

http-client/Network/HTTP/Client/Headers.hs
```diff
@@ -91,9 +91,15 @@ parseStatusHeaders mhl conn timeout' onEarlyHintHeaders cont
             Just (i, "") -> Just i
             _ -> Nothing
 
-    parseHeaders :: Int -> ([Header] -> [Header]) -> IO [Header]
-    parseHeaders 100 _ = throwHttp OverlongHeaders
+    guardMaxNumberHeaders :: Word -> IO ()
+    guardMaxNumberHeaders count =
+        when (count >= unMaxNumberHeaders mnh && count /= 0) $ do
```
tfausak, line 96: Same deal here:
```suggestion
        let n = unMaxNumberHeaders mnh
        in when (count >= n && n /= 0) $ do
```

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -820,7 +824,20 @@ data ManagerSettings = ManagerSettings
     -- Default: respect the @proxy@ value on the @Request@ itself.
     --
     -- Since 0.4.7
-    , managerMaxHeaderLength :: Maybe MaxHeaderLength
+    , managerMaxHeaderLength :: MaxHeaderLength
+    -- ^ Configure the maximum size, in bytes, of an HTTP header field.
+    -- Set it to 0 to remove this limit (eg: for debugging purposes).
```
tfausak, line 829: I don't think it's necessary to use 0 as a sentinel value. `maxBound :: Word32` is more than 4 GB, and `maxBound :: Word64` is unfathomably large. In other words, using `maxBound` would effectively disable these checks anyway.

tfausak (commented):
Looks great! I like removing the `Maybe` from both fields.

http-client/Network/HTTP/Client.hs
```diff
@@ -322,9 +323,14 @@ managerSetProxy :: ProxyOverride -> ManagerSettings -> ManagerSettings
 managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 
 -- @since 0.7.17
-managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
+managerSetMaxHeaderLength :: Word -> ManagerSettings -> ManagerSettings
```
snoyberg, line 326: This is a breaking change, I do not want to include it. It's not worth the ecosystem churn for a small improvement in type correctness.

snoyberg (changes requested):
Let's avoid any backwards-incompatible changes to the library, including to the internal modules. We can implement new features using the preferred approach though.

http-client/Network/HTTP/Client.hs
marinelli: You are right, I actually thought about it, and I didn't change it.

marinelli: But... are you talking about `managerSetMaxHeaderLength` only? I would just revert to use `Word` for both `managerSetMaxHeaderLength` and `managerSetMaxNumberHeaders`, and also the newtypes used internally for distinguishing those values.

http-client/Network/HTTP/Client/Connection.hs
marinelli: I don't know why I made this change :)

http-client/Network/HTTP/Client/Types.hs
```diff
@@ -147,12 +151,14 @@ data HttpExceptionContent
...
                    | OverlongHeaders
-                   -- ^ Either too many headers, or too many total bytes in a
-                   -- single header, were returned by the server, and the
-                   -- memory exhaustion protection in this library has kicked
-                   -- in.
+                   -- ^ Too many total bytes in a single header field were
+                   -- returned by the server.
                    --
                    -- @since 0.5.0
+                   | TooManyHeaders
```
marinelli, line 158: The introduction of this error might be considered a breaking change, but it also simplified debugging the code.

http-client/Network/HTTP/Client/Types.hs
marinelli: I agree, and this is why I just removed the `n /= 0` condition from the previous predicates.

http-client/Network/HTTP/Client.hs
marinelli: Ah, I missed your main comment above. Personally I would avoid using `Int` in one place and `Word` in an other, it would be confusing to me: why one max value is an int an an other a word?
If we want to _refine_ a bit the types of the interface, let's do that in a new version like _0.8.0_, we could put there multiple breaking changes.

snoyberg: I'm not in favor of a breaking change for this topic. I'll leave the choice to you: inconsistency in the API, or using `Int` in both places. My only stance is avoiding breaking changes.

http-client/Network/HTTP/Client.hs
```diff
@@ -324,7 +325,12 @@ managerSetProxy po = managerSetInsecureProxy po . managerSetSecureProxy po
 -- @since 0.7.17
 managerSetMaxHeaderLength :: Int -> ManagerSettings -> ManagerSettings
 managerSetMaxHeaderLength l manager = manager
-    { managerMaxHeaderLength = Just $ MaxHeaderLength l }
+    { managerMaxHeaderLength = MaxHeaderLength l }
```
snoyberg, line 328: This should also not change, yes it's an internal module, but there's no need to break end-user code for this.

http-client/Network/HTTP/Client/Types.hs
snoyberg: This is acceptable, we've documented that adding extra constructors to error types is something end users should expect.

snoyberg (changes requested):

http-conduit/test/main.hs
```diff
@@ -261,6 +262,18 @@ main = do
             let Just req1 = parseUrlThrow $ "http://127.0.0.1:" ++ show port
             _ <- httpLbs req1 manager
             return ()
+        it "too many header fields" $ tooManyHeaderFields $ \port -> do
```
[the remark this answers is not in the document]
marinelli: I moved the tests for `TooManyHeaderFields` in the `http-client` package. I tried to move also the ones for `OverlongHeaders`, without success. I didn't find the correct way to simulate it (something similar to what I did for `TooManyHeaderFields`).

http-client/Network/HTTP/Client/Connection.hs
marinelli: https://github.com/snoyberg/http-client/pull/549#discussion_r1890231380

http-client/Network/HTTP/Client/Headers.hs
marinelli: https://github.com/snoyberg/http-client/pull/549#discussion_r1890231380

snoyberg (approved):
Thanks!

Issue #426, opened 2019-12-31
Possibility of moving `mModifyRequest` and `mWrapException` to stable API?


jwoudenberg:
I'd love to be able to make on the fly changes to an existing `Manager` and wonder if it would be possible / a good idea to move (equivalents of) `mModifyRequest` and `mWrapException` to the stable api.

These are my use-cases:

For `mModifyRequest`: I'd love to put an `x-request-id` header to all outgoing HTTP requests. Such a header is sometimes used to correlate events/logs across services belonging to a single initial request from a user. In our app we read the request id from the header of an incoming request and store it in the environment for that request. For any HTTP requests made directly from the source code of the app itself we have the option of setting the `x-request-id` header directly. This isn't an option for libraries that take a `Manager` and use it to make requests on our behalf. Modifying the `Manager`'s `mModifyRequest` before passing it to such a library currently seems the best option to ensure the `x-request-id` header gets set.

For `mWrapException`: We'd love to trace all outgoing http requests. Traces consist of a start and end time of the request and are decorated with metadata from the environment the request is sent from. That metadata might be different for each outgoing request. Like the case for `mModifyRequest` we can add such tracing to requests made directly from our own source code, but to add it to requests made from libraries we need to modify the `Manager`.

snoyberg:
There are a lot of layers of crazy already in the library around modifications. I'd be worried about stabilizing what already exists, and would be more in favor of ripping off the bandage and coming up with a coherent way of dealing with all of these wrappings. One major pain point is the incoherence of modifications at Request level versus modifications at Manager level.

I'm not nixing the idea of simply stabilizing these existing things, but I'd prefer to not promote something that seems inherently broken.

jwoudenberg:
Thank you for your fast response. Also, all the best wishes for 2020!

I looked a bit at open issues and source code and noticed `requestManagerOverride` function. Is that part of pain you mean, that `Request` and `Manager` both define ways to modify the other?

Would API proposals for a more coherent modification API be a welcome or an unwelcome distraction? :)

When you talk about layers of wrappings, it make me think of the `wai` middleware design, which I quite like. Makes me what it would be like for `Manager` to be defined as a function `type Manager = RequestData -> IO (Response body)`, with modifications taking the type `Manager -> Manager`. My use cases in this issue could be implemented as a `Manager -> Manager`, as might support for automatically following redirects and proxying.

snoyberg:
[...]

PR #554, opened 2025-03-18
Make `http-client` mockable, enabling VCR-style request recording/playback

sol:
I have been using this in-house together with [assertible/vcr](https://github.com/assertible/vcr) for a couple of years now.

Getting this upstream will:

- remove the overhead of maintaining a separate fork
- allow me to use the functionality for released Haskell packages (specifically, I need this for `sol/hpack` and `hspec/sensei`)
- provide HTTP mocking and HTTP request recording/playback functionality to the wider Haskell community

Look at the [VCR README](https://github.com/assertible/vcr#readme) for how to profit from this.

There is also a [`WebMock` module](https://github.com/assertible/vcr/blob/main/src/WebMock.hs), but no documentation for it at this point.

### DIFF
--- a/http-client/Network/HTTP/Client/Core.hs
+++ b/http-client/Network/HTTP/Client/Core.hs
@@ -6,6 +6,7 @@ module Network.HTTP.Client.Core
     , httpNoBody
     , httpRaw
     , httpRaw'
+    , requestAction
     , getModifiedRequestManager
     , responseOpen
     , responseClose
@@ -25,11 +26,13 @@ import Network.HTTP.Client.Response
 import Network.HTTP.Client.Cookies
 import Data.Maybe (fromMaybe, isJust)
 import Data.Time
+import Data.IORef
 import Control.Exception
 import qualified Data.ByteString.Lazy as L
 import Data.Monoid
 import Control.Monad (void)
 import System.Timeout (timeout)
+import System.IO.Unsafe (unsafePerformIO)
 import Data.KeyedPool
 import GHC.IO.Exception (IOException(..), IOErrorType(..))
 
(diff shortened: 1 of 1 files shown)


sol:
> You are making a global hook based on `IORef`, which means it won't properly work e.g. with concurrent tests.

While this is true in general, VCR (which is built on top of this) still provides [certain concurrency guarantees](https://github.com/assertible/vcr#using-vcr-for-highly-concurrent-workloads).

> Is there any particular reason why you can't achieve the same by overriding [managerRawConnection](https://hackage.haskell.org/package/http-client-0.7.18/docs/Network-HTTP-Client.html#v:managerRawConnection) and [managerTlsConnection](https://hackage.haskell.org/package/http-client-0.7.18/docs/Network-HTTP-Client.html#v:managerTlsConnection)?

Three aspects:

1. With this approach the test code has to pass in the `Manager` explicitly.
2. Instead of working with `Request` and `Response`, the test code will have to parse requests from the wire format and render responses to the wire format.
3. I think the test code will also have to deal with connection reuse and possibly compression.

(2) and (3) are solvable engineering challenges.

Regarding (1), there is an even simpler alternative: Instead of passing in the `Manager`, pass in a function `Request -> IO Response`.

Limitations:
1. Requires you to write your code in a certain way.
1. Does not work well for acceptance tests.
1. Does not work well with third-party dependencies that do HTTP requests.

While (1) is not strictly a downside. (2) and especially (3) are real limitations.

With this PR I want to address (2) and (3).

NB: I think depending on what your requirements are, it's valid to experiment with different approaches.  The approach in this PR is simple (from a user's perspective) and it can be applied to any code, even adding tests to existing code retroactively.  But it comes at the cost of weaker concurrency guarantees. On the other hand, if you have a huge amount of tests that need to mock HTTP interactions and the execution time of your test suite becomes an issue then it might be worthwhile to pass a `makeRequest :: Request -> IO Response` function around, either explicitly or as part of some `Env` type.

@Yuras does this make sense?

Yuras:
> With this approach the test code has to pass in the Manager explicitly.

That's the whole philosophy of `http-client`. Code should allow user to configure manager one way or another, i.e. either take it from user and pass around, or use [the global one](https://hackage.haskell.org/package/http-client-tls-0.3.6.3/docs/Network-HTTP-Client-TLS.html#v:setGlobalManager)

> Instead of working with Request and Response, the test code will have to parse requests from the wire format and render responses to the wire format.

Fare enough. Then what about adding `mMakeRequest :: Request -> Manager -> IO (Response BodyReader)` to `Manager` instead of using an `IORef` for it?

sol:
> > Instead of working with Request and Response, the test code will have to parse requests from the wire format and render responses to the wire format.
> 
> Fare enough. Then what about adding `mMakeRequest :: Request -> Manager -> IO (Response BodyReader)` to `Manager` instead of using an `IORef` for it?

I think this would not address the limitations that I mentioned above:

> 1. Requires you to write your code in a certain way.
> 2. Does not work well for acceptance tests.
> 3. Does not work well with third-party dependencies that do HTTP requests.

For my specific use case this PR is appropriate.  If you (or somebody else) wants to take a different approach in the future then let's discuss this as a separate issue or PR (taking https://github.com/snoyberg/http-client/issues/426 into account).

PR #556, opened 2025-03-20
Enable caching for `stack` builds

sol:


sol:
Due to https://github.com/commercialhaskell/stack/issues/6696, this is ~2 minutes slower than what I hoped for.

PR #564, opened 2025-04-15
Propagare managerWrapException setting in mkManagerSettingsContext'

adithyaov:


adithyaov:
Ah, my bad. This was supposed to be an internal fix.
This needs an issue first to arrive on a proper fix.

PR #565, opened 2025-04-16
Propagate managerWrapException setting in mkManagerSettingsContext'

adithyaov:
`mkManagerSettingsContext'` replaces a few modifiers in the `Manager`, which prevents the `managerWrapException` from the input settings from being propagated, resulting in the loss of input behaviour.

The change in this PR uses `managerWrapException` from the input manager settings and then applies the handling function.

`mkManagerSettingsContext'` was originally idempotent; this change breaks that idempotency.

The only issue I see is that multiple applications of `mkManagerSettingsContext'` will nest the handler in `managerWrapException`. This occurs when we do something like `newTlsManagerWith tlsManagerSettings`.

**NOTE:** The same issue applies to `managerRetryableException`.

### Some context on why this is required:

We want to hook telemetry (`hs-opentelemetry`) into the HTTP manager. The ideal way to hook telemetry is to integrate it with `managerModifyRequest` and `managerModifyResponse`.

However, `managerModifyRequest` may be called multiple times, so it cannot be used for telemetry purposes. As a workaround, we use `managerWrapException`.

### DIFF
--- a/http-client-tls/Network/HTTP/Client/TLS.hs
+++ b/http-client-tls/Network/HTTP/Client/TLS.hs
@@ -97,7 +97,7 @@ mkManagerSettingsContext' set mcontext tls sockHTTP sockHTTPS = set
                 | ((fromException e)::(Maybe TLS.TLSError))==Just TLS.Error_EOF -> True
 #endif
                 | otherwise -> managerRetryableException defaultManagerSettings e
-    , managerWrapException = \req ->
+    , managerWrapException = \req act ->
         let wrapper se
               | Just (_ :: IOException)          <- fromException se = se'
               | Just (_ :: TLS.TLSException)     <- fromException se = se'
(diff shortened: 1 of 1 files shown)


sol:
Thanks for the contribution.

A couple of points in the description are a bit unclear. In particular, expressions like:

- "replaces a few modifiers"  
- "input settings"  
- "being propagated"  
- "loss of input behaviour"  

...don’t have a fixed meaning, so I can only speculate what you're referring to. You don't necessarily need to define these terms explicitly. Just make sure the issue description is consistent, clear, and brief, so that there's no doubt about what you're talking about.

Also:

1. How does this relate to https://github.com/iand675/hs-opentelemetry/tree/main/instrumentation/http-client? Are there any concrete bug reports tied to this issue?  
2. Code changes like this usually require corresponding changes to the test suite. If you believe that's not necessary in this case, please explain why.

Please **don’t reply to this comment** with answers. Instead, either **edit the existing PR description** to incorporate a clear explanation (ideally covering the above), or **close this PR and open a new one** with a clearer description.

That way, if someone looks at this PR in the future, they won’t need to scroll through a message thread to understand what it’s about. Thanks.

Issue #168, opened 2015-12-02
connectionReadLine seems wrong, or it is used incorrectly, perhaps resulting in a poor error message.


christian-marie:
I want to work out what's going on here before attempting to fix anything.

Whilst using http-client with some long-running chunked encoding GET requests, I started seeing IncompleteHeaders exceptions at a point in time when I expected headers to have long since been received.

Digging further, I came upon this in Connection.hs: 
https://github.com/snoyberg/http-client/blob/bfb355d25b209bf94a26b9cb7b60e6bca4a46b36/http-client/Network/HTTP/Client/Connection.hs#L27

```haskell
connectionReadLine :: Connection -> IO ByteString
connectionReadLine conn = do
    bs <- connectionRead conn
    when (S.null bs) $ throwIO IncompleteHeaders
    connectionReadLineWith conn bs
```

Now, in Body.hs, connectionReadLine is used like so:
https://github.com/snoyberg/http-client/blob/bfb355d25b209bf94a26b9cb7b60e6bca4a46b36/http-client/Network/HTTP/Client/Body.hs#L182

```haskell
    requireNewline = do
        bs <- connectionReadLine conn
        unless (S.null bs) $ throwIO InvalidChunkHeaders
```

Does it make sense to you why this appears to work in practice? I'm a little confused by it and have either missed something, or something is broken.

christian-marie:
Hmmn, I'm going to temporarily close this whilst I read through and understand connectionReadLineWith, which I missed.

goertzenator:
I just ran into this when dealing with a long running chunked endpoint that times out after an hour.  I agree with @christian-marie 's assessment, and will add that it doesn't matter what `connectionReadLineWith` does as the issue has manifested by then.

Please reopen this issue; I'd like to do a PR.

PR #578, opened 2026-06-02
chunked disconnect now throws InvalidChunkHeaders

goertzenator:
Fixes 11 year old issue #168

We observed a `IncompleteHeaders` exception after connecting to a streaming endpoint for an hour with a server initiated disconnect.  That is rather misleading, so this PR changes that to `InvalidChunkHeaders`.

The readline functions and call sites were also improved.

All tests pass.

### DIFF
--- a/http-client/Network/HTTP/Client/Headers.hs
+++ b/http-client/Network/HTTP/Client/Headers.hs
@@ -64,9 +64,10 @@ parseStatusHeaders mhl mnh conn timeout' onEarlyHintHeaders cont
     nextStatusLine mhl = do
         -- Ensure that there is some data coming in. If not, we want to signal
         -- this as a connection problem and not a protocol problem.
-        bs <- connectionRead conn
-        when (S.null bs) $ throwHttp NoResponseDataReceived
-        connectionReadLineWith mhl conn bs >>= parseStatus mhl 3
+        mbs <- connectionReadLineMaybe mhl conn
+        case mbs of
+            Nothing -> throwHttp NoResponseDataReceived
+            Just bs -> parseStatus mhl 3 bs
 
     parseStatus :: Maybe MaxHeaderLength -> Int -> S.ByteString -> IO (Status, HttpVersion)
     parseStatus mhl i bs | S.null bs && i > 0 = connectionReadLine mhl conn >>= parseStatus mhl (i - 1)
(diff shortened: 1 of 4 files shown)


goertzenator:
Some tests from conduit failing.  Standby...

goertzenator:
I think this PR is in good shape now.  Some CI tests are failing, but it looks like dependency rot (correct me if I'm wrong) which has nothing to do with this PR.

sol:
@goertzenator sorry for the delay, I'll try to get to this eventually.

sol:
@goertzenator I may wanna do a small refactoring before changing this code. Would you be willing to rebase your code after I do that (there will be conflicts)?

goertzenator:
For sure, go for it.

http-client/Network/HTTP/Client/Headers.hs
```diff
@@ -64,9 +64,10 @@ parseStatusHeaders mhl mnh conn timeout' onEarlyHintHeaders cont
     nextStatusLine mhl = do
         -- Ensure that there is some data coming in. If not, we want to signal
         -- this as a connection problem and not a protocol problem.
-        bs <- connectionRead conn
-        when (S.null bs) $ throwHttp NoResponseDataReceived
-        connectionReadLineWith mhl conn bs >>= parseStatus mhl 3
+        mbs <- connectionReadLineMaybe mhl conn
+        case mbs of
+            Nothing -> throwHttp NoResponseDataReceived
```
sol, line 69: I think, strictly speaking, this changes existing behavior.

Specifically, if I'm not mistaken, `NoResponseDataReceived` can now be thrown even after some data has been received (some data without CRLF, then empty chunk).

sol (changes requested):
First things first, overall, I like this change.

Your code does two things:
1. Refactoring: Effectively getting rid of `connectionReadLineWith`, which I like, as it removes some code duplication in `connectionReadLine`.  However, I'm not sure if it's possible to do that while keeping the existing behavior (see my comment on `nextStatusLine`).
2. Fix the exception semantics for chunked body readers.  You communicate the failure condition via the return value, which I think in general is preferable.  However, in this specific case we always turn `Nothing` into `throwHttp` at the call site.  So I think maybe instead pass in the error type, and keep the calls to `throwHttp` where they currently are.  I think doing so, will allow you to do (2) before (1).

To sum things up, this is what I propose:

- One PR that only does (2), addressing the issue by passing through the error type (`HttpExceptionContent`).
- If you're then still up for it, a separate PR that does (1), the refactoring (remove the code duplication in `connectionReadLine`).  In that PR we can then solely focus on what to do about `nextStatusLine`.

goertzenator:
Thanks for taking a close look at it.  I'm deep into some other stuff right now, but I hope to get back to it in within a week.

Issue #544, opened 2024-09-05
High Latency on First Request to `localhost`


julmb:
I have this code:
```hs
main :: IO ()
main = do
    request <- parseRequest "http://localhost:3000/test"
    getCurrentTime >>= print
    httpBS request >>= T.putStr . T.decodeUtf8Lenient . getResponseBody
    getCurrentTime >>= print
    httpBS request >>= T.putStr . T.decodeUtf8Lenient . getResponseBody
    getCurrentTime >>= print
```
With a simple `servant` service at `http://localhost:3000/test`, I get:
```
2024-09-05 03:47:47.018037777 UTC
Hello, world!
2024-09-05 03:47:47.271253156 UTC
Hello, world!
2024-09-05 03:47:47.272148168 UTC
```

Given that this is all local, I would have expected these requests to finish very fast. However, the first request takes about 250ms, only subsequent ones are fast.

I did some investigating, and it seems that `getAddrInfo` returns `[::1]:3000` and `127.0.0.1:3000`, in that order. The function `firstSuccessful` then attempts to connect to each of these, staggered by 250ms, which only succeeds for the second address.

Not sure if there is an easy/elegant solution to this, but it is very unfortunate that a simple connection to `localhost` incurs such a high latency on the first request.

julmb:
For what it's worth, this is how `curl` handles a connection to `localhost:3000`:
```
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* connect to ::1 port 3000 from ::1 port 49152 failed: Connection refused
*   Trying 127.0.0.1:3000...
* Connected to localhost (127.0.0.1) port 3000
```
It looks like it also tries to connect to the IPv6 address first, but instead of waiting 250ms, it seems to immediately try the IPv4 address once the connection is refused by the IPv6 address. The whole exchange takes less than 5ms.

From the comment above the function `firstSuccessful`, it seems like it is using the staggered approach to comply with [RFC 8305](https://datatracker.ietf.org/doc/html/rfc8305). I am not sure if trying the next address immediately after the previous address has been refused would also be compliant or a good idea.

PR #579, opened 2026-06-03
Don't let crypton-connection do name resolution

### LINKED ISSUE #569: IPv6/IPv4 and 'Happy Eyeballs'
There have been [some accumulating issues](https://github.com/commercialhaskell/stack/issues/5994) at Stack's repository that appear to be IPv6-related for Linux, macOS and Windows users.

Somebody who knows much more about this area than me has observed that, basically, the lower level is doing a lookup, getting an IPv6 address, and trying to connect to it despite IPv6 not being configured either locally or on the router; or the reverse (which might happen in some situations), where the lookup gets an IPv4 address and tries to connect to it over an IPv6-only connection.

For its part, the `curl` project (held up as an exemplar) [explains](https://everything.curl.dev/transfers/conn/how.html#how-libcurl-connects):

> How libcurl connects
> When libcurl is about to do an Internet transfer, it first resolves the host name to get a number of IP addresses for the host. A hostname needs to have at least one address for libcurl to be able to connect to it.
> 
> A hostname can have both IPv4 addresses and IPv6 addresses and they can have a set of both.
> 
> If the host only returned addresses of a single IP family, libcurl iterates over each address and tries to connect. If the connect attempt fails for an IP, libcurl continues to try the next entry until the entire list is exhausted. ...
> 
> Happy Eyeballs
> When it has received both IPv4 and IPv6 addresses for a host, libcurl first tries to connect to an IPv6 address and after a short delay it tries connecting to the first IPv4 address - at the same time and in parallel. Once one of the attempts succeeds, the others are discarded. This method of attempting to connect using both families at the same time is called Happy Eyeballs and is the widely accepted best practice for Internet clients. ...

For its part, ChatGPT offers up (rightly or wrongly) that 'performing a DNS resolution (which returns both IPv4 and IPv6 addresses) and then attempting connections sequentially in the order returned can result in bad user experience if, say, IPv6 is preferred but broken'.

Does the `http-client` family of packages do anything similar to 'Happy Eyeballs'? If not, should they or could they?

### LINKED ISSUE #5994: connection failure
I am on arch linux and I installed stack through ghcup. However, running stack always resutlts in a connection failure.

### Steps to reproduce

Run command `stack update`

### Expected

It should run without errors.

### Actual

~~~text
stack update --verbose
Version 2.9.1, Git revision 409d56031b4240221d656db09b2ba476fe6bb5b1 x86_64 hpack-0.35.0
2022-12-28 10:20:54.442871: [debug] Checking for project config at: /home/mechap/stack.yaml
2022-12-28 10:20:54.443596: [debug] Checking for project config at: /home/stack.yaml
2022-12-28 10:20:54.443634: [debug] Checking for project config at: /stack.yaml
2022-12-28 10:20:54.443661: [debug] No project config file found, using defaults.
2022-12-28 10:20:54.454144: [debug] (SQL) SELECT COUNT(*) FROM "last_performed" WHERE ("action"=?) AND ("timestamp">=?); [PersistInt64 1,PersistUTCTime 2022-12-27 09:20:54.453673226 UTC]
2022-12-28 10:20:54.456163: [info] Selected mirror https://hackage.haskell.org/
2022-12-28 10:20:54.456258: [info] Downloading root
2022-12-28 10:20:59.463542: [error] HttpExceptionRequest Request {
  host                 = "hackage.haskell.org"
  port                 = 443
  secure               = True
  requestHeaders       = [("Accept-Encoding",""),("User-Agent","Haskell pantry package")]
  path                 = "/root.json"
  queryString          = ""
  method               = "GET"
  rawBody              = False
  redirectCount        = 10
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = 0.0.0.0:0, addrCanonName = Nothing}, host name: Just "hackage.haskell.org", service name: Just "443"): does not exist (Try again))
~~~

~~~text
curl -I https://hackage.haskell.org:443/root.json
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 4216
Server: nginx/1.18.0 (Ubuntu)
Content-Type: application/json
Cache-Control: public, no-transform, max-age=60
Content-MD5: MKHLRXfk9A1kOc9X0OA7Jg==
ETag: "30a1cb4577e4f40d6439cf57d0e03b26"
Accept-Ranges: bytes
Date: Wed, 28 Dec 2022 09:24:43 GMT
Via: 1.1 varnish
Age: 0
X-Served-By: cache-cdg20728-CDG
X-Cache: HIT
X-Cache-Hits: 1
X-Timer: S1672219483.965029,VS0,VE459
Vary: Accept-Encoding
~~~

### Stack version

~~~text
stack --version
Version 2.9.1, Git revision 409d56031b4240221d656db09b2ba476fe6bb5b1 x86_64 hpack-0.35.0
~~~

Yuras:
Instead use the existing resolver.

Fixes #569 <s>and #544</s> (EDIT: no, it doesn't), likely fixes https://github.com/commercialhaskell/stack/issues/5994

The problems is that `crypton-connection` tries addresses [one by one](https://github.com/kazu-yamamoto/crypton-connection/blob/1af72aad1733046195e6f0025adcb7944fa75dad/Network/Connection.hs#L243-L252), so if the fist address returned by `getaddrinfo(3)` is a misconfigured ipv6 address, it'll spend quite a bit of time trying to connect to ipv6 address, and only then try the next address.

Note that `http-client` tries connections [concurrently](https://github.com/snoyberg/http-client/blob/5f0d14724af440e439ce5803b941911609e766e8/http-client/Network/HTTP/Client/Connection.hs#L218-L227).

I tested only `getTlsConnection` part and not ` getTlsProxyConnection` since I don't actually have any proxy on hands. I let maintainer decide what to do next.

### DIFF
--- a/http-client-tls/Network/HTTP/Client/TLS.hs
+++ b/http-client-tls/Network/HTTP/Client/TLS.hs
@@ -132,8 +132,6 @@ getTlsConnection mcontext tls sock = do
               , NC.connectionUseSocks = sock
               }
         withSocket (const $ pure ()) ha host port $ \socket -> do
-            -- This block is exception-safe thanks to withSocket.
-            -- We won't send TLS bye in case of exception, but that's OK
             conn <- NC.connectFromSocket context socket params
             convertConnection conn
 
@@ -156,8 +154,6 @@ getTlsProxyConnection mcontext tls sock = do
                           $ fromIntegral port
               }
         withSocket (const $ pure ()) ha host port $ \socket -> do
-            -- This block is exception-safe thanks to withSocket.
-            -- We won't send TLS bye in case of exception, but that's OK
             conn <- NC.connectFromSocket context socket params
             NC.connectionPut conn connstr
             conn' <- convertConnection conn
--- a/http-client-tls/Network/HTTP/Client/TLS.hs
+++ b/http-client-tls/Network/HTTP/Client/TLS.hs
@@ -124,15 +124,18 @@ getTlsConnection :: Maybe NC.ConnectionContext
                  -> IO (Maybe HostAddress -> String -> Int -> IO Connection)
 getTlsConnection mcontext tls sock = do
     context <- maybe NC.initConnectionContext return mcontext
-    return $ \_ha host port -> bracketOnError
-        (NC.connectTo context NC.ConnectionParams
-            { NC.connectionHostname = strippedHostName host
-            , NC.connectionPort = fromIntegral port
-            , NC.connectionUseSecure = tls
-            , NC.connectionUseSocks = sock
-            })
-        NC.connectionClose
-        convertConnection
+    return $ \ha host port -> do
+        let params = NC.ConnectionParams
+              { NC.connectionHostname = strippedHostName host
+              , NC.connectionPort = fromIntegral port
+              , NC.connectionUseSecure = tls
+              , NC.connectionUseSocks = sock
+              }
+        withSocket (const $ pure ()) ha host port $ \socket -> do
+            -- This block is exception-safe thanks to withSocket.
+            -- We won't send TLS bye in case of exception, but that's OK
+            conn <- NC.connectFromSocket context socket params
+            convertConnection conn
 
 getTlsProxyConnection
     :: Maybe NC.ConnectionContext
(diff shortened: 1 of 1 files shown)


sol:
@Yuras do you think it's possible to first do a small refactoring that extracts the `bracketOnError` stuff into a small helper function, and then, in a second commit, apply the semantic change at one place?

1. one commit that extracts the common pattern into a helper function and does not introduce any behavioral changes
2. a second commit that introduce the semantic change

Yuras:
Not sure what exactly you mean, `withSocket` seems to be exactly the helper you are talking about... If you want to do it differently, you can just close the PR.

sol:
> Not sure what exactly you mean, `withSocket` seems to be exactly the helper you are talking about... If you want to do it differently, you can just close the PR.

This looks like a useful contribution, so I hope there is a better outcome than "close the PR".

That said:
- Doing the refactoring first makes it easier for me to review.
- There's also some code duplication.  For example if source code comments are necessary, I would rather have them in one place.
- I'm not doing this myself.  I just spent a lot of time trying to fix CI.  One reason I did this is so that I can review your PR, and so that we can get it merged.

sol:
Also, note that I was asking you a question. That means that I'm not even sure if this turns out nice.  If I already know 100% that something is the right thing to do, then I will not phrase it as a question, but instead say "please do X".

Yuras:
Looks like there is a misunderstanding here. I'm happy to refactor the PR if you explain what exactly you are expecting. But I genuinely don't understand what I should do. And since the changes are quite trivial, it's almost definitely be faster for you to do the changes than to explain it to me.

> There's also some code duplication. For example if source code comments are necessary, I would rather have them in one place.

The comments are probably not necessary, I'm happy to drop them. It was just me being confused by the design of `withSocket` - if you compare it with e.g. `System.IO.withFile`, you'd expect it to close the socket on exit, but it does that only on exception.

Anyway, I used `withSocket` because that's how [http-client](https://github.com/snoyberg/http-client/blob/d717e7566802586b6f0b9e0dfd4815cf2a20c39f/http-client/Network/HTTP/Client/Connection.hs#L162-L163) and [http-client-openssl](https://github.com/snoyberg/http-client/blob/d717e7566802586b6f0b9e0dfd4815cf2a20c39f/http-client-openssl/Network/HTTP/Client/OpenSSL.hs#L42-L58) are implemented.

I don't see any code duplication here. The `ConnectionParams` are different, the last argument of `withSocket` is also different in both cases. There is literally nothing to deduplicate, unless you mean refactoring `withSocket` itself, which is much more intrusive. But, again, I'm happy to deduplicate the code if you point me to the exact thing to deduplicate.

sol:
> Looks like there is a misunderstanding here. I'm happy to refactor the PR if you explain what exactly you are expecting.

OK, at least we sorted that out.  That's the refactoring I had in mind https://github.com/snoyberg/http-client/pull/584.

The second commit of https://github.com/snoyberg/http-client/compare/tls-no-resolve then introduces your change.

sol:
> I'm happy to deduplicate the code if you point me to the exact thing to deduplicate.

It's less about deduplication, but more about making it easier to review, especially as I also don't have a way to test the proxy code path.

Yuras:
Cool, thanks. Closing in favor to your branch.

PR #581, opened 2026-06-06
Fix the build

sol:


sol:
@snoyberg I would like to make the following adjustments to the repository settings:

- required that CI passes before merging via branch protection rules
- enable "Allow auto-merge"
- disable "Allow merge commits"

I would eventually also want to automate Hackage releases, but this will require a little bit more work, as my current setup would only work if the whole family of packages would be released with the same version numbers.

Can you give me the required permissions, or do you want me to provide exact steps on what to change?

snoyberg:
I don't think I can give you those permissions since this is on my personal account, not an org. If you give me the steps you'd like me to take, I'll make the updates.

sol:
> I don't think I can give you those permissions since this is on my personal account, not an org. If you give me the steps you'd like me to take, I'll make the updates.

Thanks @snoyberg.  I'll open a separate issue for this once CI is green.

sol:
The Window build is still flaky, but I'm not in a position to debug this.

PR #585, opened 2026-06-07
Don't let crypton-connection resolve hostname 

### LINKED ISSUE #569: IPv6/IPv4 and 'Happy Eyeballs'

### LINKED ISSUE #5994: connection failure

sol:
Instead use the existing resolver.

@Yuras said in #579:

Fixes #569 <s>and #544</s> (EDIT: no, it doesn't), likely fixes https://github.com/commercialhaskell/stack/issues/5994

The problems is that `crypton-connection` tries addresses [one by one](https://github.com/kazu-yamamoto/crypton-connection/blob/1af72aad1733046195e6f0025adcb7944fa75dad/Network/Connection.hs#L243-L252), so if the fist address returned by `getaddrinfo(3)` is a misconfigured ipv6 address, it'll spend quite a bit of time trying to connect to ipv6 address, and only then try the next address.

Note that `http-client` tries connections [concurrently](https://github.com/snoyberg/http-client/blob/5f0d14724af440e439ce5803b941911609e766e8/http-client/Network/HTTP/Client/Connection.hs#L218-L227).

I tested only `getTlsConnection` part and not ` getTlsProxyConnection` since I don't actually have any proxy on hands. I let maintainer decide what to do next.

### DIFF
--- a/http-client-tls/Network/HTTP/Client/TLS.hs
+++ b/http-client-tls/Network/HTTP/Client/TLS.hs
@@ -125,14 +125,14 @@ getTlsConnection :: Maybe NC.ConnectionContext
                  -> IO (Maybe HostAddress -> String -> Int -> IO Connection)
 getTlsConnection mcontext tls sock = do
     context <- maybe NC.initConnectionContext return mcontext
-    return $ \_ha host port -> do
+    return $ \ha host port -> do
         let params = NC.ConnectionParams
               { NC.connectionHostname = strippedHostName host
               , NC.connectionPort = fromIntegral port
               , NC.connectionUseSecure = tls
               , NC.connectionUseSocks = sock
               }
-        withConnection context params
+        withConnection context params ha host port
             convertConnection
 
 getTlsProxyConnection
@@ -142,7 +142,7 @@ getTlsProxyConnection
     -> IO (S.ByteString -> (Connection -> IO ()) -> String -> Maybe HostAddress -> String -> Int -> IO Connection)
 getTlsProxyConnection mcontext tls sock = do
     context <- maybe NC.initConnectionContext return mcontext
-    return $ \connstr checkConn serverName _ha host port -> do
+    return $ \connstr checkConn serverName ha host port -> do
         let params = NC.ConnectionParams
               { NC.connectionHostname = strippedHostName serverName
               , NC.connectionPort = fromIntegral port
@@ -152,7 +152,7 @@ getTlsProxyConnection mcontext tls sock = do
                       Just _ -> error "Cannot use SOCKS and TLS proxying together"
                       Nothing -> Just $ NC.OtherProxy (strippedHostName host) $ fromIntegral port
               }
-        withConnection context params $ \conn -> do
+        withConnection context params ha host port $ \conn -> do
             NC.connectionPut conn connstr
             conn' <- convertConnection conn
 
@@ -162,8 +162,9 @@ getTlsProxyConnection mcontext tls sock = do
 
             return conn'
 
-withConnection :: NC.ConnectionContext -> NC.ConnectionParams -> (NC.Connection -> IO a) -> IO a
-withConnection context params = bracketOnError (NC.connectTo context params) NC.connectionClose
+withConnection :: NC.ConnectionContext -> NC.ConnectionParams -> Maybe HostAddress -> String -> Int -> (NC.Connection -> IO a) -> IO a
+withConnection context params ha host port action = withSocket (const $ pure ()) ha host port $ \socket -> do
+  NC.connectFromSocket context socket params >>= action
 
 convertConnection :: NC.Connection -> IO Connection
 convertConnection conn = makeConnection


http-client-tls/Network/HTTP/Client/TLS.hs
```diff
@@ -162,8 +162,9 @@ getTlsProxyConnection mcontext tls sock = do
 
             return conn'
 
-withConnection :: NC.ConnectionContext -> NC.ConnectionParams -> (NC.Connection -> IO a) -> IO a
-withConnection context params = bracketOnError (NC.connectTo context params) NC.connectionClose
+withConnection :: NC.ConnectionContext -> NC.ConnectionParams -> Maybe HostAddress -> String -> Int -> (NC.Connection -> IO a) -> IO a
+withConnection context params ha host port action = withSocket (const $ pure ()) ha host port $ \socket -> do
+  NC.connectFromSocket context socket params >>= action
```
sol, line 167: @Yuras regarding your original comment:

```haskell
-- We won't send TLS bye in case of exception, but that's OK
```

That relates to `NC.connectionClose`, I guess.  So I assume we don't properly terminate the connection, but we close the socket, which is technically all we care about.

Comparing `connectTo` to `withSocket`, there seems to be a lot of things going on in `connectTo`, including things related to SOCKS.  So I can't trivially convince myself that this won't break the proxy code path (+ apparently that code path is also not covered by the test suite).  I'll either have to look at it again when I got more sleep, or somebody has to test it. 😅

https://github.com/kazu-yamamoto/crypton-connection/blob/1af72aad1733046195e6f0025adcb7944fa75dad/Network/Connection.hs#L219

sol:
Another question is, whether we would want to address this in `crypton-connection` instead, not sure?

@kazu-yamamoto

Yuras:
> That relates to NC.connectionClose, I guess. So I assume we don't properly terminate the connection, but we close the socket, which is technically all we care about.

This is true. Though one can argue we are relying on the current implementation of `crypton-connection`. Nothing stops `connectFromSocket` from allocating resources meant to be deallocating in `connectionClose`. Note that `connectFromSocket` already calls `Network.TLS.contextNew` when `connectionUseSecure` is set (not our case fortunately.)

> So I can't trivially convince myself that this won't break the proxy code path

Fair. Though `connectTo` only looks complicated, we call it with `OtherProxy`, in which case it's only doing name resolution and calls `connectFromSocket`. But I understand your concern.

Yuras:
> whether we would want to address this in crypton-connection instead,

It definitely should be fixed in `crypton-connection` as well. But IMO it doesn't make much sense for `http-client` to use two different name resolutions. Since we already have our own implementation, then let's use it everywhere.

kazu-yamamoto:
@sol I forked `connection` to `crypton-connection` just to solve the dependency problem and don't know the internal at all. So, I cannot answer your question.

@Yuras If you will send a PR to `crypton-connection`, I would love to merge it.

sol:
> > So I can't trivially convince myself that this won't break the proxy code path
> 
> Fair. Though `connectTo` only looks complicated, we call it with `OtherProxy`, in which case it's only doing name resolution and calls `connectFromSocket`.

Yeah, that makes sense.  Thanks for pointing this out. Given that we always pass in `OtherProxy`, most of that code is not relevant.

Overall, this looks good to me, and I'm leaning towards merging it.

Looking at it in detail, this is what I understand.

Both code paths, old and new:
- connect to the same `host` / `port` (especially for `getTlsProxyConnection` we still use the correct host)
- use `strippedHostName` (but admittedly, it's a little bit messy to not have it done at one single place)

Differences:
- The old code uses `AI_ADDRCONFIG`, which is probably a good idea for the sequential approach, but shouldn't make much of a difference for the async approach.
- exception semantic changes (e.g. `error` instead of `HostNotResolved`)
- The new code path uses `TCP_NODELAY`.
- the new code takes the optional `HostAddress` (`ha`) into account (which unconditionally  uses IPv6)

Regarding `TCP_NODELAY`, to understand whether this is a good or a bad idea in general, I would need to look at the individual places where we send data.  But that's not relevant here. What we need to ask for this PR: Is there any reason why we would want to use a different setting for TLS connections.  If (after the TLS handshake) every "plaintext send" results in exactly one send on the socket, then I would say using the same `TCP_NODELAY` setting as for plaintext connections is probably the right thing to do.  If that's indeed the case, I don't know.  I assume yes, but I didn't check.

Regarding `HostAddress`, I don't know, would need to look at the call sites.  I didn't do that.

@Yuras any thoughts?

Yuras:
> @Yuras any thoughts?

I don't have any opinion regarding `AI_ADDRCONFIG` and `TCP_NODELAY`. Non-secure connections worked so far, I don't see why these options won't work for secure connection as well.

Regarding `HostAddress`. Indeed this changes the behavior, though IMO it doesn't matter. I thinks host address was meant to be used, it was just not implemented. E.g. `opensslManagerSettings` uses host address in `managerTlsConnection`. Note that host address in `managerTlsProxyConnection` is [never used](https://github.com/snoyberg/http-client/blob/05c45f9680180181763a399625cca48c5c0a638e/http-client/Network/HTTP/Client/Manager.hs#L273)

Yuras:
> @Yuras If you will send a PR to crypton-connection, I would love to merge it.

Sorry, but I personally don't use `crypton-connection` for anything. So as long as `http-client` is fixed, I'm good.

sol:
Thanks @Yuras.

I'm reasonably confident that:
- `AI_ADDRCONFIG` should be fine with "Happy Eyeballs"
- `TCP_NODELAY` at worst changes performance characteristics

If you are reasonably confident about the PR as a whole, to the level that you would click the merge button yourself, then I'm OK with merging this.

If so, then please give me a final 👍

Also, would it be OK if I ask you to open a separate PR that:
- bumps the version (major)
- adds a changelog entry that references this PR
- (fixes ...) in the commit message, so that the corresponding issue gets closed

If that's OK with you?  I didn't plan to take over this PR, I'm not at my computer right now, and I have plenty of other things on my plate 😅

